PR #802, opened 2015-12-10
add error handling so that sfx outage will fail gracefully fixes #666

### LINKED ISSUE #666: Error Handling for SFX Outage
The same type of "generic error" message is displayed for ejournal records when the SFX API is unavailable.  Is there a way to improve this to still display the record but indicate the access is temporarily unavailable.

pgwillia:
also included configurable timeout value which requires app restart.


redlibrarian:
@pgwillia yay - thanks very much!

Issue #1001, opened 2016-05-19
Summary of Symphony library code changes


redlibrarian:
Libraries removed:

AITF_CALG - Alberta Innovates - Technology Futures - Calgary
AHS_QEII - Alberta Health Services - Queen Elizabeth II Hospital
AHS_WHC - Alberta Health Services - Women's Health Centre
AHS_PLC - Alberta Health Services - Peter Lougheed Centre

Library added:

AHS_INT - Alberta Health Services - Internet
Last week we changed the description to:       Alberta Health Services/Covenant Health - Internet
This items belonging to AHS_INT need to be returned in Alberta Health Services and Covenant Health search results. I don't know if you have any filters on searches.

redlibrarian:
I just realized why that facet value remained "ahsint" - it's because the data needs to be riendexed. It should be fine when we deploy.

PR #1019, opened 2016-07-14
 fixing issue: Summary of Symphony library code changes #1001

astrilet:
Before:
screenshot 2016-07-14 12 41 41
After:
screenshot from 2016-07-14 11 48 04


redlibrarian:
I just realized why that facet value remained "ahsint" - it's because the data needs to be riendexed. It should be fine when we deploy.

Issue #856, opened 2016-01-12
word wrap problem for filters list in Firefox


theLinkResolver:
![line breaks](https://cloud.githubusercontent.com/assets/14913579/12271847/7ae7e570-b91a-11e5-8880-84004add01f6.jpg)

Browser window is at full width.

Chrome handles word wrap better, for the most part.

![chrome](https://cloud.githubusercontent.com/assets/14913579/12271961/feaf02bc-b91a-11e5-9482-094f8fb0d648.jpg)

redlibrarian:
@nnunn Do you have any thoughts about this? Is this a Bootstrap thing?

theLinkResolver:
Much improved

Issue #983, opened 2016-04-29
Articles: delete Collection from facet list


redlibrarian:

redlibrarian:
This will have to wait until our access to Ebsco admin is restored.

seanluyk:
Collection facet is no longer appearing, closing

Issue #762, opened 2015-12-01
Fix OUR license width in ejournal display.


redlibrarian:
Example record: https://search.library.ualberta.ca/catalog/954925545272

theLinkResolver:
@redlibrarian These still don't look right when the browser window is small. (example from Firefox)

![our width](https://cloud.githubusercontent.com/assets/14913579/13576913/2891d5ac-e44d-11e5-9b78-d671860cd7c7.jpg)

redlibrarian:
OK. Not sure what I'll be able to do with this now that our design person is away.

theLinkResolver:
Fixed in test.

PR #1023, opened 2016-07-20
fixing issue #856

astrilet:
modified css file to take effect for firefox browser only

### DIFF
--- a/app/assets/javascripts/ualib-discovery.js
+++ b/app/assets/javascripts/ualib-discovery.js
@@ -13,3 +13,13 @@ $(document).ready(function () {
 	});
 	$(".staff").stupidtable();
 });
+
+function setiFrameAttr() {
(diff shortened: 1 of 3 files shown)


astrilet:
After fixing:
screenshot from 2016-07-20 13 08 03

astrilet:
this also includes fix to issue #762 
screenshot from 2016-07-21 13 24 12

astrilet:
removed collection from facet list, issue #983. Noticed that "Publication Year" appears twice (see screenshot attached) I modified one that we are generating explicitly with label "Publication Year 888"(this modification is not in the code that I checked in), another one comes form the search results. Do we need to change that? 
screenshot from 2016-07-22 14 44 27

redlibrarian:
Merged into development - closing PR.

PR #1133, opened 2017-10-10
Profiles password

nmacgreg:
(Re: my CR684)
I've moved the credentials for editing Profiles, into secrets.yml. 
Please review!


nnunn:
When I try to run this on my dev machine I get a type error - I think it's a encryption problem. I'll try and fix it.

nnunn:
Ok all fixed - I did change  the variables to match ruby standards so  profiles_edit_user and  profiles_edit_password. I also had to take out the full path from the secret.yaml file. You will need to do the same in the production secrets file...

Issue #1144, opened 2018-05-05
Infrastructure and Workflows


redlibrarian:
@pgwillia to
- [x] resurect Weiwei's docker branch
- [x] Move Solr config back into Discovery repo
- [x] add some bits to the Readme about how to use it
- [x] orient Sam and Natasha to using docker in development

@redlibrarian to
- [x] document discovery servers and their role
- [ ] Flesh out issues (bugs, user stories, etc)
- [x] security updates - Deal with Gemnasium warnings (Dependabot?)
- [x] Set up GitHub workflows
- [x] Request UAT server from ITS (or find out if current DI UAT will serve for Discovery)

Then we can talk about next steps
- [x] tests (start fresh with minitest?)
- [x] style (rubocop/hound)
- [x] travis-ci
- [x] docker UAT environment

redlibrarian:
@weiweishi Is there anything else you need in here for now?

pgwillia:
In order to setup Solr for Discovery in a Docker contianer, I'll be copying [blacklight_solr_conf](https://github.com/ualbertalib/blacklight_solr_conf) into discovery.  There are some notes in that's repo's readme that describe why Neil wants them separate. He should be in the loop about this change.

pgwillia:
The security updates I was thinking about are the Gemnasium warnings.

redlibrarian:
OK, I've fleshed out the Discovery server information as best I can. I have one question for Neil, which I will get the answer for and then update the file.

PR #1150, opened 2018-05-07
Docker for development

### LINKED ISSUE #1148: Move Solr config back into Discovery repo
When we moved to SolrCloud, I moved the Solr config to its [own repo](https://github.com/ualbertalib/blacklight_solr_conf). To get the repository as close as possible to Jupiter (to work best with docker, CI, and UAT), we want the solr configuration to be put back into the Discovery repository. I'll also make sure that the directory structure matches what's in Jupiter.

- [ ] Move solr config directories back into ualbertalib/discovery
- [ ] Adjust solr config directories to match those in ualbertalib/jupiter

pgwillia:
- [x] update README.md
- [x] create docker-compose.lightweight.yml which will provide backing datastores for mysql and solr
- [x] ensure that the solr container is working closes #1148
- [x] ensure that the mysql container is working
- [x] db:setup
- [x] ingest some data
- [x] make sure the results of localhost:3000/catalog are sane
- [x] upgrade migrations so that Ruby 2.4 can be used by all

Part of #1144


config/solr.yml
```diff
@@ -1,19 +1,13 @@
...
-  url: <%= Rails.application.config.solr %>
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/development' %>
+test:
+  url: <%= ENV['SOLR_TEST_URL'] || 'http://localhost:8983/solr/test' %>
+uat:
+  <<: *default
+staging:
+  <<: *default
 production:
-  url: <%= Rails.application.config.solr %>
```
pgwillia: This doesn't need to change.  I'll keep it in line with the existing way.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ruby:2.3.4
+MAINTAINER wshi
```
murny: Maintainer is deprecated, should use a label like here: https://github.com/ualbertalib/jupiter/blob/master/Dockerfile.deployment#L2

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+  mysql:
+    image: mysql:5.7
+    environment:
+      MYSQL_ROOT_PASSWORD: mysecretpassword
+    volumes:
+      - mysql:/var/lib/mysql
+
+  solr:
+    image: solr:6.0
+    # Uncomment this if you want to access SOLR's web interface
```
murny: Can remove this comment now since you have this already uncommented?

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ruby:2.3.4
```
murny: README says we need Ruby 2.4? But here its Ruby 2.3? Should this be updated to reflect the README?

Gemfile
```diff
@@ -2,9 +2,8 @@ source 'http://rubygems.org'
 
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '4.2.10'
+gem 'rails'
```
murny: You going to have a bad time if you don't lock this down? This will jump to the latest rails version, currently rails 5.2 which I am guessing will break this application?

docker-compose.lightweight.yml
```diff
@@ -0,0 +1,30 @@
...
+version: '3'
+
+volumes:
+  mysql:
+    driver: local
+  solr:
+    driver: local
+
+services:
+  mysql:
```
murny: Most developers will have a mysql service running in the background. This will fail as a result because of using the same port mappings. 

Would maybe suggest either leave this commented out with instructions so developers can enable if needed or change the default port below or put a comment about this issue so people can resolve this themselves (maybe in the readme)?

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+MAINTAINER wshi
+
+# Need to add jessie-backports repo so we can get FFMPEG, doesn't come with jessie debian by default
```
murny: Comments probably not applicable in discovery as I doubt they need FFMPEG

PR #1152, opened 2018-05-08
Fixes #1128.

### LINKED ISSUE #1128: display UA shield when holdings include item in NEOS_FREE library
([transferred from helpdesk](https://helpdesk.library.ualberta.ca/otrs/index.pl?Action=AgentTicketZoom;TicketID=41412#))

Sample records:
https://www.library.ualberta.ca/symphony?q=7616648
https://www.library.ualberta.ca/symphony?q=Supreme%20Court%20Law%20Review (first result)

(Sort of related to #339?)

redlibrarian:
Added NEOS Free to the list of locations recognized as UAL.


pgwillia:
You could move all the logic for displaying the UAL shield to a helper
```
def display_ual_shield(document)
  # all the reasons you'd want to display the ual_shield for this document
end
```

Then the [view](https://github.com/redlibrarian/discovery/blob/335b386f66b42a7597f1bbfce29fc4bf27cb0431/app/views/catalog/_index_default.html.erb#L23-L30) would just be
```
<!-- Display UAL shield -->
<% if display_ual_shield(document) %> 
        <img alt="licensed for U of A" title="licensed for U of A" src="//s3.amazonaws.com/libapps/sites/165/icons/6665/ualib-logo.png" style="height:8%;width:8%"/>.
<% end %>

And changing the logic for when to display the shield might become easier to do and easier to read.

pgwillia:
@redlibrarian Just wondering what the status of this one is?

redlibrarian:
@pgwillia You're obviously right about the wisdom of breaking that logic out into a helper - there's a lot of cleanup like that that we should do on the views. I think at this point, to get this changes into production, I'd like to just proceed with it as it is.

pgwillia:
@redlibrarian cool.  I'm happy to review. There were some additions to this PR.  Could you talk me through what you changed and why? I see a whole file being deleted and seems to be replaced by a txt file -- what did this do? Is it documentation? There's a small change to a docker-compose.yml too which was removed from the code base. It's causing conflicts, perhaps you could remove this change?

redlibrarian:
@pgwillia I don't think I made the change to docker-compose... I'll see if I can find out from the history. And I don't know what library_codes.txt would have been (especially a binary file). All our config files have been yaml. Clearly I still need to learn how to manage commits/PRs more cleanly.

redlibrarian:
@pgwillia OK; those changes should now be reverted.

redlibrarian:
Since these are all data-related, this goes back to my overall question about how best to test these kinds of changes. My suspicion is that we need to build some kind of test dataset that we thoroughly "know" in order to test these, but I'm not really sure. 

The third one is difficult to test because we read that in in real-time via the API. So do we have to mock the API response?

pgwillia:
I wasn't even talking about formal/automated testing, but that's a good idea too. Even in a development environment what can I visually inspect that will verify that these changes are working? I'm mostly blocked by not being familiar with the interface/content.

redlibrarian:
Ah, I understand, sorry. With these kinds of issues, I usually look for records that will show the desired outcome. (Ideally these would be part of some kind of standard test dataset). These can be tricky, and sometimes I have to try a few things, or ask bib services for example records. In the case of the NEOS_FREE Shield piece, the issue #1128 has a record which should be fixed by adding the NEOS_FREE logic (https://www.library.ualberta.ca/symphony?q=7616648), so I would look at that record, comparing before and after adding the logic in dev.

For the RCRF one, I would do a search and find the existing label, first in the library facet, then in any of the records that facet retusn (Research Collection Resource Facility) and make sure it updates to the new one (Research and Collections Resource Facility).

For the third one, again, I'd probably have to get a record from bib services to test. With this one, we're adding a display label for a new status which won't be in the index until the next full reindex (which took place this morning and isn't in production yet).

pgwillia:
Could you merge the latest from ualbertalib:master into your local master prior to merging? I think this is fine to merge but just to be sure I'd feel more comfortable if both branches had the same history.

redlibrarian:
@pgwillia I merged ualbertalib:master and pushed everything up. I think that should be OK, if you want to give it one more look.

PR #1169, opened 2018-05-23
ruby 2.1.5 compatible Gemfile.lock

pgwillia:
resolves issue with ruby 2.1.5 build created here: https://github.com/ualbertalib/discovery/commit/bacfd27191ee1b2de150e698af384d6e0524b4b6#diff-e79a60dc6b85309ae70a6ea8261eaf95

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -73,13 +73,13 @@ GEM
       sass (>= 3.3.4)
     bootstrap_form (2.7.0)
     builder (3.2.3)
-    capybara (3.0.3)
+    capybara (2.18.0)
       addressable
       mini_mime (>= 0.1.3)
-      nokogiri (~> 1.8)
-      rack (>= 1.6.0)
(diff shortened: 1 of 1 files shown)


nnunn:
I think this looks good, but honestly I'm not completely sure of the issue.

Gemfile.lock
```diff
@@ -350,8 +350,8 @@ GEM
     warden (1.2.7)
       rack (>= 1.0)
     xml-simple (1.1.5)
-    xpath (3.0.0)
```
pgwillia: The automated build uses Ruby version 2.1.5 so this line causes
`Gem::InstallError: xpath requires Ruby version >= 2.2.
An error occurred while installing xpath (3.0.0), and Bundler cannot continue.
Make sure that `gem install xpath -v '3.0.0'` succeeds before bundling.`

Using the older version of the gem is compatible with Ruby 2.1.5.

PR #1176, opened 2018-06-08
Add 926l mapping for collections.

redlibrarian:


config/SolrMarc/symphony_index.properties
```diff
@@ -142,6 +142,7 @@ source_tesim = "Symphony"
 holdings_tesim = 949[a-z]
 owning_library_tesim = 999m
 #location_tesim = 926m
+collection_tesim = 926l
```
pgwillia: I'm trying to understand what's happening here so that I might be able to make similar changes in the future.  I can see that this file is used to ["define environment-specific config files" for SolrMarc](https://github.com/ualbertalib/discovery/blob/a2a81ce5da20041386fd02009e4d1255d93d1e7c/config/SolrMarc/symphony_config.properties#L1-L4). 

I recognize *_tesim as the english text, stored, indexed and multivalued dynamic field in Solr.  It's not obvious to me if or where this is used in the discovery application. Is it? What is collection_tesim?

Googling for documentation of MARC I found that [9xx is for local implementation](https://www.loc.gov/marc/96principl.html#six).  What does 926l mean, and is it documented anywhere?

I found a sample in the fixtures to test this: spec/fixtures/marc_display_with_print_holdings.xml. It looks like it will pull out "ON_SHELF". Is that the desired behaviour? 
```
<datafield tag="926" ind1=" " ind2=" "><subfield code="a">HT 151 H82 1973</subfield><subfield code="w">LC</subfield><subfield code="c">1</subfield><subfield code="i">000043591213</subfield><subfield code="d">1/2/2001</subfield><subfield code="l">ON_SHELF</subfield><subfield code="m">UABARD</subfield><subfield code="n">1</subfield><subfield code="p">$150.00</subfield><subfield code="r">Y</subfield><subfield code="s">Y</subfield><subfield code="t">BOOK</subfield><subfield code="u">10/25/1993</subfield><subfield code="o">.STAFF. A0074591</subfield></datafield>
```
But how would I run this?  `rake ingest[symphony]` looks like it only pulls ["data/sample.mrc"](https://github.com/ualbertalib/discovery/blob/076c5a2f39caabe412716a9946e2028331dbb131/config/ingest.yml#L61)

Thanks in advance for providing some clarity.

redlibrarian:
This came out of the NEOS Blacklight project, which uses the Discovery index. In the current NEOS Catalogue, we have an advanced search limit called 'location' which uses a particular **holdings**-level status field to allow users to prelimit a search to a particular collection (e.g. Easy Reading or Curriculum):

screen shot 2018-06-12 at 8 01 03 am

We don't have access to the **holdings** in the Solr index, as we only read them via the API for real-time display. However, there's a **bib**-level field (926 subfield l) which contains the (I think) same information. In order to build a facet that replicates the "location" prelimit, we need to index the 926l field.

The symphony_index.properties file is where all the MARC field mapping is specified for the Symphony data (there's another one for SFX which also uses MARC). I added the 926l field so that we have it in the index and then I can see about building a facet with it.

"ON_SHELF" is one of the locations/collections - the data is very messy, but that's one possible value. I'm not sure how best to test individual data points like that, but I thought the test would likely be with building the facet, as this change will only get the 926l into the Solr index. 

I hope that helps clarify some things. I realize I move fast and don't do much documentation with these kinds of changes, because the whole thing is in my head, and I'm not used to working with others on this project. I'll try to change that.

pgwillia:
Out of curiousity what happens when a mapping fails? I'm wondering what happens with missing data, multiple values, incorrect mappings, are there other cases?

redlibrarian:
@pgwillia The data in the catalogue is often a mess, and because it's all strings its difficulty to verify/test, except empirically after the fact, by having people report problem records. Missing data is sometimes legitimate (many records won't have a given field), incorrect mappings is fairly subjective and relies on user or cataloguing department feedback. How the system handles multiple values is determined by the schema.xml file specifying whether a solr field can be multivalued or not (though other ways of dealing with multivalued fields are handled in the more sophisticated mappings of the BeanShell scripts [here](https://github.com/ualbertalib/discovery/tree/master/config/SolrMarc/index_scripts) ). There are likely many other cases where mapping can "go wrong", but they're generally subjective; the best thing is for the mapping system to silently accomodate whatever data is provided (because the data is not very clean) and then rely on user-reports or cataloguing to determine problems.

redlibrarian:
After discussing this in the NEOS Blacklight group, the 926$l is not an appropriate field to use for this functionality. This is going back to the drawing board.

PR #1183, opened 2018-06-22
Introducing RAILS_ENV staging

pgwillia:
I am specifically not trying to change how anything currently works with this PR. I believe this is the minimum that needs to be done to introduce a staging environment to replace the use of RAILS_ENV test in our current deployments.


config/database.yml
```diff
@@ -21,6 +21,14 @@ default: &default
   pool: 5
   timeout: 5000
   encoding: utf8
+  
+mysql: &mysql
```
pgwillia: I guess this is strictly not required, but it seems strange to me that the config sets up sqlite when mysql is used everywhere else.

pgwillia: I guess this is strictly not required because it is created from a template in ansible, but it seems strange to me that the config sets up sqlite when mysql is used everywhere else.

config/environments/staging.rb
```diff
@@ -0,0 +1,85 @@
+Rails.application.configure do
```
pgwillia: copy of the production environment

config/logger.yml
```diff
@@ -4,6 +4,9 @@ development:
 test:
   log_path: "log/ingest.log"
 
+staging:
```
pgwillia: think it makes more sense to use the production example than what was set for test here.

config/secrets.yml
```diff
@@ -25,6 +25,10 @@ development:
 test:
   secret_key_base: 0bc7ee59d6ffbd6a29779c50fc7dad3f01a7cc18292634e4852cf1b2adeecf77f577b0cb46f6145a9da62d60903d8b9ab2d310f9b2e5f3883d19b630c26b02d6
 
+staging:
+  secret_key_base: 0bc7ee59d6ffbd6a29779c50fc7dad3f01a7cc18292634e4852cf1b2adeecf77f577b0cb46f6145a9da62d60903d8b9ab2d310f9b2e5f3883d19b630c26b02d6
```
pgwillia: config/secrets.yml is actually created from a template so this isn't strictly necessary either.

config/solr.yml
```diff
@@ -13,7 +13,9 @@
 
 development:
   url: "http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test"
-test: &test
-  url: <%= Rails.application.config.solr %>
+test:
+  url: <%= ENV['DISCOVERY_SOLR_TEST_URL'] || 'http://localhost:8983/solr/discovery-test' %>
+staging:
+  url: <%= Rails.application.config.solr %>
```
pgwillia: This value is set in application.rb using value at ingest.yml. Ingest.yml is again, is created from a template in ansible.

pgwillia:
```
diff --git a/projects/discovery/roles/discovery/tasks/test.yml b/projects/discovery/roles/discovery/tasks/test.yml
index e4d1d30..f3a9387 100644
--- a/projects/discovery/roles/discovery/tasks/test.yml
+++ b/projects/discovery/roles/discovery/tasks/test.yml
@@ -2,8 +2,8 @@
 
   - name: set environment variables based on server role test
     set_fact:
-        passengerAppEnv: "test"
-        railsEnv: "test"
+        passengerAppEnv: "staging"
+        railsEnv: "staging"
         serverName: "search-test.library.ualberta.ca"
 
 # Create a file on the server that will set environment variables key to troubleshooting
diff --git a/projects/discovery/roles/discovery/templates/database.yml.test.j2 b/projects/discovery/roles/discovery/templates/database.yml.test.j2
index 9796528..4585966 100644
--- a/projects/discovery/roles/discovery/templates/database.yml.test.j2
+++ b/projects/discovery/roles/discovery/templates/database.yml.test.j2
@@ -3,7 +3,7 @@
 # The destination file for this template is: /var/www/sites/blacklight/config/database.yml
 
 # Note, the serverRole=uat maps here, too
-test:
+staging:
   adapter: mysql2
diff --git a/projects/discovery/roles/discovery/templates/secrets.yml.j2 b/projects/discovery/roles/discovery/templates/secrets.yml.j2
index 829fa93..0bcf54d 100755
--- a/projects/discovery/roles/discovery/templates/secrets.yml.j2
+++ b/projects/discovery/roles/discovery/templates/secrets.yml.j2
@@ -17,7 +17,7 @@
-test:
+staging:
   secret_key_base: 
```

nmacgreg:
The Unix team is struggling with larger issues than code, in deciding how to handle this new environment.  For instance, we have an entire VLAN set aside for Dev, and multiple for Test, each with infrastructure, although much flattened in comparison with Prod.  For instance, should we create a VLAN tree mimicking Prod c/w firewall rules, for Staging?   Should we have a separate instance of Nagios, dedicated to Staging? How about other infrastructure: Gluster, oVirt, MySQL, PostgreSQL, haproxy? Nested oVirts? Do we need additional hardware? Or is this just ansible-config/inventory-staging/, and otherwise sharing our Test environment?   The playbooks for multiple applications are built around a "server-role" environment variable, that sometimes even surfaces as an environment-variable in custom install scripts inside an RPM, so this is a thorny issue.  We'll try to hammer some of these issues flat, in our team meeting today.  We can't really deal with code, until these decisions are made.

config/database.yml
nmacgreg: Originally, the Dev environment was a VM, with a checked-out copy of this repo as a working environment.  Sam felt a real copy of MySQL was too much work, and overkill for development, so sqllite was his choice.

config/logger.yml
nmacgreg: Thanks, makes sense.  We do occasionally perform ingests in the Staging environment, especially when the Solr schema is changed.

Issue #1164, opened 2018-05-17
Developer Happiness


pgwillia:
I [reverted changes](https://github.com/ualbertalib/discovery/pull/1163) that might cause an issue with immediate release. These will be addressed in a branch separate from master until a path forward can be worked out

- [x] readme to orient to the project, maybe even some resources to understand catalog concepts
- [x] green test suite -- licence to make changes
- [x] consistent style -- [bbatsov's rubocop](https://github.com/bbatsov/rubocop)
- [x] security concerns addressed
- [x] ~~recent ruby~~
- [x] debugging tools --  better errors
- [x] address deprecation warnings
- [x] Docker
  - [x] lightweight datastores for development
  - [x] travis-ci needs some way of running solr to run tests
  - [x] UAT - needs datastores, rails, nginx, some way to do ingestion

PR #1189, opened 2018-07-06
Dockerized Discovery

pgwillia:
Proof of concept to dockerize Discovery so that features in the master branch can be deployed in a UAT environment for review.  #1164 

There are some awkward things here that may be cleaned up.

- [x] May conflict with the activities of renaming 'test' servers to 'staging'.


Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM phusion/passenger-ruby21:0.9.34
```
pgwillia: Chose this image because it supports older ruby (2.1.5 as required by the project) also uses nginx+passenger.  Passenger is used in the way that this app is deployed in production so maintain that similarity.

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+CMD ["/sbin/my_init"]
+
+RUN bash -lc 'rvm install ruby-2.1.5'
+RUN bash -lc 'rvm --default use ruby-2.1.5'
+RUN bash -lc 'gem install bundler'
+
+RUN apt-get update -qq \
+    && apt-get install -y build-essential \
+                          mysql-client \
+                          default-jre \
```
pgwillia: Required by SolrMarc for ingest tasks, but the config/ingest.yml is not configured here.

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+                          imagemagick \
```
pgwillia: Required by the paperclip gem.

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    && rm -rf /var/lib/apt/lists/*
+    
+ENV APP_ROOT /app
+RUN mkdir -p $APP_ROOT
+WORKDIR $APP_ROOT
+
+# setup nginx/passenger
+RUN rm -f /etc/nginx/sites-enabled/default
+COPY config/nginx.conf /etc/nginx/sites-enabled/discovery.conf
+COPY config/rails-env.conf /etc/nginx/main.d/rails-env.conf
```
pgwillia: I'm not convinced this is the right way to do this.  But the documentation claims the env vars aren't passed by ngnix unless they're explicitly stated https://github.com/phusion/passenger-docker#setting-environment-variables-in-nginx

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Preinstall gems in an earlier layer so we don't reinstall every time any file changes.
+COPY Gemfile  $APP_ROOT
+COPY Gemfile.lock $APP_ROOT
+RUN bundle install --without development test --jobs=3 --retry=3
+
+# *NOW* we copy the codebase in
+COPY --chown=app:app . $APP_ROOT
+
+# Precompile Rails assets.
+RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
```
pgwillia: This seems to change every time.  Is there a better way?

bin/setup
```diff
@@ -19,6 +19,9 @@ Dir.chdir APP_ROOT do
 
   puts "\n== Preparing database =="
   system "bin/rake db:setup"
+  
+  puts "\n== Preparing the CMS =="
+  system "bin/rake comfortable_mexican_sofa:cmssetup"
```
pgwillia: Might as well add this to the setup script.

config/nginx.conf
```diff
@@ -0,0 +1,18 @@
+# /etc/nginx/sites-enabled/discovery.conf:
+server {
+    listen 80;
+    server_name uat.library.ualberta.ca;
```
pgwillia: Couldn't figure out how to use variables here.

config/rails-env.conf
```diff
@@ -0,0 +1,7 @@
+# /etc/nginx/main.d/rails-env.conf:
```
pgwillia: Basically a copy of the .env_deployment_sample but with the format that nginx expects.

docker-compose.lightweight.yml
```diff
@@ -0,0 +1,28 @@
...
+  mysql:
+    image: mysql:5.7
+    environment:
+      MYSQL_ROOT_PASSWORD: mysecretpassword
+    volumes:
+      - mysql:/var/lib/mysql
+    ports:
+      - "3306:3306"
+
+  solr:
```
pgwillia: Potentially useful for developers in testing.  The development and uat image points at production like data, not this.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    image: ualbertalib/discovery:uat
+    depends_on:
+      - mysql
+    env_file: .env_deployment
+    volumes:
+      - assets:/app/public/assets/
+      - file-storage:/app/storage/
+    restart: always
+    ports:
+      - "3002:80"
```
pgwillia: We decided in coordination with other projects on uat that this one would be available on port 3002.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+  mysql:
+    restart: always
+    image: mysql:5.7
+    env_file: .env_deployment
+    volumes:
+      - mysql:/var/lib/mysql
+
+  # Rails
+  web:
+    image: ualbertalib/discovery:uat
```
pgwillia: When/if this is merged this should change to ualbertalib/discovery and the docker hub build settings should be modified.  In the meantime occasional re-bases of master on this branch will build new images.

.env_deployment_sample
```diff
@@ -0,0 +1,14 @@
+RAILS_ENV=uat
+DATABASE_URL=mysql2://root:mysecretpassword@mysql/discovery_uat
+SOLR_URL=http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test
```
pgwillia: This is the same value that development is currently set to. This feels awkward to me, but one of the pain points with developing this application is the data problems that arise. Since this side of the application is read-only I don't see any problem with this other than publishing something that external parties won't have access to.  Write to the application is configured conf/ingest.yml.

pgwillia: Just found out that uat doesn't have access to this 
```
App 352 output: Errno::ETIMEDOUT (Connection timed out - connect(2) for "solrcloud-test.library.ualberta.ca" port 8080):
```

config/rails-env.conf
```diff
@@ -1,7 +1,8 @@
 # /etc/nginx/main.d/rails-env.conf:
 env RAILS_ENV=uat;
 env DATABASE_URL=mysql2://root:mysecretpassword@mysql/discovery_uat;
-env SOLR_URL=http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test;
+env SOLR_URL=http://solr:8983/solr/discovery;
```
pgwillia: Discovered uat server can't reach ual solrcloud-test

Dockerfile
murny: Is there a specific reason discovery is using ruby 2.1.5? Ruby 2.1 is End of Life and no longer supported

pgwillia: Yes.  In conversation with Neil we determined that updating Ruby for this project was a significant amount of work which we decided wasn't appropriate use of time for our "band-aid summer project."

murny: But what's preventing it from being upgraded if anything? I bet moving to ruby 2.3/2.4 etc would take hardly any work at all? and would make things simpler in the long run.

No different then the Solr version issue previously? Solr was thought it needed a really old version so we ended up spending a ton of time back porting and creating a docker image to support this old version and then after all this work, we realised we can just use a latest version of Solr and use the communities Solr image.

pgwillia: This is all the work required development side.  https://github.com/ualbertalib/discovery/commit/23be61676a1419f73057b1bd13088d827a718468

app/helpers/blacklight/articles_helper_behavior.rb
```diff
@@ -70,11 +70,15 @@ def eds_connect
 
     # creates EDS API connection object, initializing it with application login credentials
     @connection = ConnectionHandler.new(2)
-    File.open(auth_file_location,"r") {|f|
-        @api_userid = f.readline.strip
-        @api_password = f.readline.strip
-        @api_profile = f.readline.strip
-    }
+    if ENV['EDS_API_PROFILE']
+      @api_profile = ENV['EDS_API_PROFILE']
```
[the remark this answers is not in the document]
pgwillia: Will be glad when this is gone!

pgwillia:
I've rebased on the current master.  This would be okay to merge in my absence if you think it moves the project forward.  There shouldn't be any changes needed to configs for deployment, but there may be some extra files that the spec file that builds the rpm would want to exclude.

.env_deployment_sample
```diff
@@ -0,0 +1,13 @@
+RAILS_ENV=uat
+DATABASE_URL=mysql2://root:mysecretpassword@mysql/discovery_uat
+SOLR_URL=http://solr:8983/solr/discovery
+SECRET_KEY_BASE=cf7098b3a239fa2741c0e88cd7a2b18ba1b732179c0fc85865e31fbd80dbc1b2a03f201af3165d9fae52296731041f1eb721b2e9cc547fd92e527039c7d472f2
+# EDS info to create APIauthentication.txt
+EDS_API_USER_ID=
```
murny: Seems like these, `EDS_API_USER_ID/EDS_API_PASSWORD` are never used? (nothing parses them, only EDS_API_PROFILE is being used and pretty much all you need) Are they needed?

config/ingest.yml
```diff
@@ -5,7 +5,7 @@
 # If you're making substantive changes that need to reach Test & Prod, please also 
 # make those changes to ansible's copies of this file!
 
-solr: http://localhost:8983/solr
+solr: http://solr:8983/solr/discovery
```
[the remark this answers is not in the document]
murny: Will this happen in this PR regarding making this a ENV Var instead? Or another etc?

config/rails-env.conf
murny: Does NGINX actually need any of these ENV Vars?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    depends_on:
+      - mysql
+    env_file: .env_deployment
+    volumes:
+      - assets:/app/public/assets/
+      - file-storage:/app/storage/
```
murny: Is discovery using `/app/storage/`? Assume this be for ActiveStorage but doubt discovery has this?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    restart: always
+    image: ualbertalib/blacklight_solr_conf
+    volumes:
+      - solr:/opt/solr/server/solr/mycores
+
+  # Rails
+  web:
+    image: ualbertalib/discovery:uat
+    depends_on:
+      - mysql
```
murny: Should `web` container also depend on the `solr` one?

config/rails-env.conf
pgwillia: Yes. Passenger needs these ENV vars to configure the rails application, so they have to be passed through nginx. 

This is the problem in production where we compromised on using config/secrets.yml rather than ENV vars because the application couldn't read the ENV vars because of [this quirk of Passenger](https://www.phusionpassenger.com/library/indepth/environment_variables.html#passenger-served-apps).

config/ingest.yml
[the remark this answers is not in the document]
pgwillia: It's in another: https://github.com/ualbertalib/discovery/pull/1227
That's ready to go now too.  Feel free to review that next.

docker-compose.yml
pgwillia: Probably no.  Copied from Jupiter.  I'll double check and remove if not.

.env_deployment_sample
pgwillia: Strictly no. EDS is going away soon too.  Might as well not add this.  Thanks!

docker-compose.yml
pgwillia: Yep!

PR #1192, opened 2018-07-24
WIP - refactor ingest to use solr defined by RAILS_ENV instead of static value

pgwillia:
*  add lightweight docker containers with README instructions
*  introduces writeable property in solr.yml to protect read-only intended access

Use caution when merging this PR. This will definitely change the way the application is deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Integration tests (run against <http://search-test.library.ualberta.ca/>)
 ## Additional Tasks for uat, staging and production
 
 1.  `bundle exec rake assets:precompile` this can take several minutes
-2.  `comfortable_mexican_sofa:cmssetup` to setup the CMS
+2.  `bundle exec rake comfortable_mexican_sofa:cmssetup` to setup the CMS
 3.  create cron jobs to ingest `bundle exec rake ingest[sfx]`, `bundle exec rake ingest[databases]` and clean session table `bundle exec rake sessions:cleanup`
 
 ## Ingest


config/application.rb
```diff
@@ -23,7 +23,6 @@ class Application < Rails::Application
 
     @c = YAML.load_file('config/ingest.yml') rescue {}
     config.proxy = @c['proxy']
-    config.solr = @c['solr']
```
pgwillia: This is where the url for solr is set currently in test/production. I want to move this back into the configuration rather than making the config/ingest.yml brittle and needing to be tracked in both this repo and ansible.

config/ingest.yml
```diff
@@ -1,12 +1,3 @@
-# *** WARNING *** 
-# Ansible contains customized copies of this file, for Test & Prod environments.
-# Why? Specifically, to address the changing URL for solr, and Neil's inability to
-# use lineinfile to achieve the desired results.
-# If you're making substantive changes that need to reach Test & Prod, please also 
-# make those changes to ansible's copies of this file!
-
-solr: http://localhost:8983/solr
```
pgwillia: This is where the url for solr is set currently in test/production. I want to move this back into the configuration rather than making the config/ingest.yml brittle and needing to be tracked in both this repo and ansible.

config/solr.yml
```diff
@@ -10,10 +10,15 @@
 # jetty_path is only required for rake tasks that need to know
 # how to start up solr, generally for automated testing.
 
-
+default: &default
+  url: <%= Rails.application.secrets.solr_url %>
```
pgwillia: Move to using ENV vars to set the values with reasonable defaults. For deployment ease, have one place where all these can be set.

config/solr.yml
```diff
@@ -10,10 +10,15 @@
...
 development:
-  url: "http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test"
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/discovery' %>
 test: &test
-  url: <%= Rails.application.config.solr %>
+  url: http://localhost:8983/solr/discovery-test
+  writable: true
+ingest: 
+  <<: *default
+  writable: true
```
pgwillia: Started with a read-only attribute but decided that the default should be read-only and to guard unless this writable attribute is set. Also introduced an ingest RAILS_ENV so that ingest and production (or development, or staging, etc) could occur on the same application server.

lib/tasks/ingest_test_data.rake
```diff
@@ -1,11 +0,0 @@
-require_relative "./ingest_test_data_helper"
-
-include RakeTasks
-
-desc 'ingest test data'
```
pgwillia: Doesn't seem to be working?

lib/tasks/ingest_test_data_helper.rb
```diff
@@ -1,34 +0,0 @@
...
-  def ingest_database_test_set
-    args = {:collection => "database_test_set"}
-    Rake.application['ingest'].execute(Rake::TaskArguments.new(args.keys, args.values))
-  end
-
-  def ingest_symphony_test_set
-    `rake ingest[symphony_test_set]`
-  end
-
-  def ingest_all_test_sets
```
pgwillia: Doesn't seem to be working?

README.md
```diff
@@ -6,18 +6,21 @@ discovery platform. Based on [Project Blacklight](projectblacklight.org).
...
-2.  run `bundle install`
-3.  run `bundle exec rake db:setup`
-4.  Create the APIAuthentication.txt (containing your EBSCO credentials) and token.txt (empty) files
-5.  `bundle exec rails s`
-6.  Point your browser to **<http://localhost:3000/catalog>** or **<http://localhost:3000/results>**
+2.  run `docker-compose -f docker-compose.lightweight.yml up -d`
+3.  run `bundle install`
+4.  run `bundle exec rake db:setup`
+5.  Create the APIAuthentication.txt (containing your EBSCO credentials) and token.txt (empty) files
+6.  `RAILS_ENV=ingest bundle exec rake ingest[database_test_set] && RAILS_ENV=ingest bundle exec rake ingest[sfx_test_set] && RAILS_ENV=ingest bundle exec rake ingest[symphony_test_set]`
```
pgwillia: The the cron jobs would be configured to run like this as well `RAILS_ENV=ingest bundle exec rake ingest[databases]` etc

lib/ingest/ingester.rb
```diff
@@ -10,7 +10,7 @@ def initialize
     log_dir = "#{Rails.root}/#{log_dir}"
     FileUtils.mkdir_p(log_dir) unless File.directory?(log_dir)
     log_file = File.open(log_dir+"/"+log_file, File::WRONLY|File::APPEND|File::CREAT)
-    @@ingest_log = Logger.new(log_file)
```
pgwillia: Is this necessary?  Giving `warning: class variable access from toplevel`

redlibrarian:
@pgwillia This looks like a good solution to me; I didn't know it was possible to set read-only/writable in the Solr configuration.

lib/ingest/ingester.rb
redlibrarian: I implemented Logger to add logging to specific parts of the application. It was ages ago, and I'm not sure those logs (as opposed to the application logs) were ever looked at. Removing it will only stop logging of the few ingest errors it was set up to catch. I imagine there are better logging solutions than this.

pgwillia:
`config/blacklight.yml` is just a file so you can put whatever you want in it. I added logic to the rake tasks before write and delete to abort if this isn't set so that we only purposefully write to Solr targets we intend to.

lib/ingest/ingester.rb
pgwillia: Specifically does the @@ingest_log need to be a class variable?

redlibrarian: IIRC the Logger documentation used a class variable so there was only a single instance of the Logger gathering/writing data. If there's another way to do that, then it shouldn't need to be a class variable.

PR #1195, opened 2018-07-25
Discovery: refactor ingest to use solr defined by RAILS_ENV instead of static value

pgwillia:
*  add lightweight docker containers with README instructions 
*  introduces writeable property in solr.yml to protect read-only intended access

Use caution when merging this PR. This will definitely change the way the application is deployed.


README.md
```diff
@@ -39,14 +42,14 @@ Integration tests (run against <http://search-test.library.ualberta.ca/>)
 
 1.  `bundle exec rake assets:precompile` this can take several minutes
 2.  `bundle exec rake comfortable_mexican_sofa:cmssetup` to setup the CMS
-3.  create cron jobs to ingest `bundle exec rake ingest[sfx]`, `bundle exec rake ingest[databases]` and clean session table `bundle exec rake sessions:cleanup`
+3.  create cron jobs to ingest `RAILS_ENV=ingest bundle exec rake ingest[sfx]`, `RAILS_ENV=ingest bundle exec rake ingest[databases]` and clean session table `bundle exec rake sessions:cleanup`
```
pgwillia: @nmacgreg note this change!

config/application.rb
```diff
@@ -23,7 +23,6 @@ class Application < Rails::Application
 
     @c = YAML.load_file('config/ingest.yml') rescue {}
     config.proxy = @c['proxy']
-    config.solr = @c['solr']
```
pgwillia: This is where the url for solr is set currently in test/production. I want to move this back into the configuration rather than making the config/ingest.yml brittle and needing to be tracked in both this repo and ansible.

config/blacklight.yml
```diff
@@ -10,10 +10,15 @@
 # jetty_path is only required for rake tasks that need to know
 # how to start up solr, generally for automated testing.
 
-
+default: &default
+  url: <%= Rails.application.secrets.solr_url %>
```
pgwillia: Move to using ENV vars to set the values with reasonable defaults. For deployment ease, have one place where all these can be set.

config/blacklight.yml
```diff
@@ -10,10 +10,15 @@
...
 development:
-  url: "http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test"
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/discovery' %>
 test: &test
-  url: <%= Rails.application.config.solr %>
+  url: http://localhost:8983/solr/discovery-test
+  writable: true
+ingest: 
+  <<: *default
+  writable: true
```
pgwillia: Started with a read-only attribute but decided that the default should be read-only and to guard unless this writable attribute is set. Also introduced an ingest RAILS_ENV so that ingest and production (or development, or staging, etc) could occur on the same application server.

config/blacklight.yml
```diff
@@ -10,10 +10,15 @@
 # jetty_path is only required for rake tasks that need to know
 # how to start up solr, generally for automated testing.
```
pgwillia: Renamed file because of deprecation warning.

config/ingest.yml
```diff
@@ -1,12 +1,3 @@
-# *** WARNING *** 
-# Ansible contains customized copies of this file, for Test & Prod environments.
-# Why? Specifically, to address the changing URL for solr, and Neil's inability to
-# use lineinfile to achieve the desired results.
-# If you're making substantive changes that need to reach Test & Prod, please also 
-# make those changes to ansible's copies of this file!
-
-solr: http://localhost:8983/solr
```
pgwillia: This is where the url for solr is set currently in test/production. I want to move this back into the configuration rather than making the config/ingest.yml brittle and needing to be tracked in both this repo and ansible.

pgwillia:
Made a little tweak to this and github closed #1192. Trying again.

@nmacgreg @redlibrarian could you both look at this and let me know if this might be an okay way to handle read/write separation to solr targets? These changes anticipate reclaiming the test RAILS_ENV. Testing is really hard to do until this is complete.

pgwillia:
Revising with @nmacgreg feedback

Issue #1103, opened 2017-03-08
Discovery: Complete test suite


redlibrarian:
Once the work on issue #1144 is done, we can start thinking about how to tackle testing. We may want to start with a couple of very high-level "sanity check" tests, and then add more granular tests as we go. This might also mean removing a lot of the specs written for various unit and integration tests. The dev team has adopted minitest as their test framework, which is fine with me. We can add thoughts about how to approach / get on top of tests to this issue as we discuss. Setting up a couple of high-level tests *that pass* is essential for completing the GitHub workflows (#1147). Currently, tests fail, which means requiring status checks won't work.

redlibrarian:
- [ ] Flesh out integration and acceptance tests.
- [ ] Unit tests for new gems

redlibrarian:
If, as per #940 , we port our customizations over to the latest version of Blacklight, then we can write tests as we port the code. This is done when:
- [ ] Unit tests for Helpers
- [ ] Unit tests for Services
- [ ] Integration
- [ ] Acceptance

tests are completed.

redlibrarian:
Verify the docker-compose
Session with Tricia on workflow
Investigate Gemnasium
Investigate Dependabot
Setup Github Workflow
Move Solr config back into Discovery repo
Set up UAT server
Decision on how to approach tests
Create working test suite

(Clean up this issue).

pgwillia:
So tests are [green now](https://travis-ci.org/ualbertalib/discovery) but coverage is poor:'

Coverage report generated for RSpec to /home/travis/build/ualbertalib/discovery/coverage. 688 / 2180 LOC (31.56%) covered.

The [grab bag tests are mostly passing](http://cardiff.library.ualberta.ca/job/discovery-grabBag/lastBuild/console) (just one known issue causing failure -- @nmacgreg was going to look at it.)

I can look at adding a test like this one to the spec to try to improve coverage.
t/dataSanityRegression.t ... 
ok 1 - search-test.library.ualberta.ca: Search for shakespeare
ok 2 - Did we get content?
ok 3 - Check for masked error
ok 4 - Results contain a count of the number of books
ok 5 - Extraction of count successful
ok 6 - Count of books should exceed 14,000, actual: 19019

Did you have others in mind?

pgwillia:
[...]

PR #1172, opened 2018-06-05
Green tests

pgwillia:
This will remove the failing unit/feature tests and provide some license to make changes.  We can then remove the Jenkins job and rely on Travis/Github to keep us honest.
#1164

PR #1197, opened 2018-07-26
Discovery: more tests please

pgwillia:
#1103 

With Solr datastore available and ingest tasks run for the test sets, more tests pass. `Coverage report generated for RSpec to ./discovery/coverage. 981 / 2231 LOC (43.97%) covered.`

Depends on #1195


.travis.yml
```diff
@@ -1,9 +1,22 @@
 language: ruby
+addons:
+  chrome: stable
```
murny: Not sure what the default `sudo` is? 

But if your adding chrome addon to travisci and getting random failure errors, you need to enforce`sudo: required` in your travisci

https://docs.travis-ci.com/user/chrome#selecting-a-chrome-version

More info here:
https://github.com/travis-ci/travis-ci/issues/8836

config/blacklight.yml
```diff
@@ -10,10 +10,15 @@
...
 # how to start up solr, generally for automated testing.
 
-
+default: &default
+  url: <%= Rails.application.secrets.solr_url %>
 development:
-  url: "http://solrcloud-test.library.ualberta.ca:8080/solr/discovery-test"
-test: &test
-  url: <%= Rails.application.config.solr %>
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/discovery' %>
```
murny: Maybe you can consolidate the config here? You have some in ENV vars, some in this file, some in application secrets?

Could just put everything in application secrets?

Than this file just becomes:

```
default: &default
  url: <%= Rails.application.secrets.solr_url %>
development: 
  <<: *default
test: 
  <<: *default
  writable: true
ingest: 
  <<: *default
  writable: true
production: 
  <<: *default
```
etc.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe "delete" do
+    before do
+      @controller.send(:current_or_guest_user).save
+      @controller.send(:current_or_guest_user).bookmarks.create! document_id: '2007020969', document_type: "SolrDocument"
+    end
+    
+    it "has a 200 status code when delete is success" do
+      xhr :delete, :destroy, :id => '2007020969', :format => :js
+      expect(response).to be_success
+      expect(response.code).to eq "200"
```
murny: Could be wrong, but i think these two assertions are exactly the same?
```
expect(response).to be_success
expect(response.code).to eq "200"
```

`be_success` checks if the response was a 200 (compared to a 302 (redirect), 404 (not found) or 5xx (error))


One of the biggest issues with rspec is that you can do the same thing 50 different ways, but only one way is deemed "the right way" by the community.

In RSpec world, i believe the best practice for checking responses is to use `have_http_status` so for example you can use a number code, symbol or generic code (e.g. :error is 5xx)


```
expect(response).to have_http_status(200)
expect(response).to have_http_status(:success)
expect(response).to have_http_status(:ok)
```
More info: https://relishapp.com/rspec/rspec-rails/docs/matchers/have-http-status-matcher

spec/controllers/databases_controller_spec.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "../rails_helper"
+
+describe DatabasesController, type: :controller do
+
+  it "should have a collection name" do
+    get :index
+    expect(assigns(:collection_name)).to eq "Databases"
```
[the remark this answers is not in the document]
pgwillia: I agree that the tests could be improved a lot.  A bunch of the test suite that I inherited with the project was failing.  My goal with this is that 
* now that the 'test' environment has been reclaimed 
* a Solr docker container with the blacklight_solr_conf is available as a datastore that travis can run
* ingest target is configurable
* tests helper runs a ingest of the sample marc
-> the tests that I had previously removed (#1172) now pass.

pgwillia:
Ingest is getting a bit of a rewrite.  Will incorporate the feedback here in the new PR.

PR #1200, opened 2018-08-03
Discovery: Library facet alpha sort

### LINKED ISSUE #1178: Discovery: "Library" facet should sort alphabetically
See #857 for previous discussion.

**Description**:
The library facet doesn't quite sort alphabetically. This can be seen most clearly on the Advanced Search page (https://library.ualberta.ca/advanced) where *all* the libraries appear in the pre-search filter:

screenshot from 2018-06-12 10-14-36

**Priority**: High

**Expected Behavior**:
Library names are sorted alphabetically with their display name. 

**Diagnosis**:
The reason this is happening is because the values of the facet are Symphony library codes, not library names. We map the code to the display name in the view. So for example, the reason **Alberta Government Library - Internet** comes before **Alberta Government Library - Capital Boulevard** is because the facet values are actually AGINTERNET and AGL_CAP respectively. 

**Proposed solution**
I think the correct solution for this is to do the same thing we do with the **institution** facet, and map the display name to codes during the indexing process. It looks like I already created a BeanShell script to do this, so it may be that the only thing to do is to using [location_facet.bsh](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/index_scripts/location_facet.bsh) intead of locations.bsh in the [Symphony indexing rules](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/symphony_index.properties).

**Block?**
If we move to Traject (#1138) for indexing, then we'll need to replicate all the MARC mapping in traject, so there's no point in implementing this *solution* in SolrMarc. If we do the Traject work first, then we should make sure the codes are mapped to the display names in Traject from the beginning.

I propose we park the Traject work for now, and focusing on fix this in currently Blacklight. (Weiwei)

pgwillia:
fixes #1178 

depends on #1197 and #1195


app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -124,12 +124,7 @@ def facet_partial_name(display_facet = nil)
   # @option options [Boolean] :suppress_link display the facet, but don't link to it
   # @return [String]
   def render_facet_value(facet_solr_field, item, options ={})
-    if facet_solr_field == "location_tesim"
-      path = search_action_path(add_facet_params_and_redirect(facet_solr_field, item))
-      content_tag(:span, :class => "facet-label") do
-        link_to_unless(options[:suppress_link], facet_display_value(facet_solr_field, @locations[item.value] ), path, :class=>"facet_select")
-      end + render_facet_count(item.hits)
-    elsif facet_solr_field == "languagenote_tesim"
+    if facet_solr_field == "languagenote_tesim"
```
pgwillia: Mapping the location from it's code to its description is no longer required by the application.

config/SolrMarc/index_scripts/location_facet.bsh
```diff
@@ -8,87 +8,77 @@ public Set getLocationFacet(Record record){
...
+  location_codes.put("uamath", "University of Alberta Mathematics");
+  location_codes.put("uascitech", "University of Alberta Cameron - Science & Technology");
+  location_codes.put("uasjc",  "University of Alberta St Joseph's College");
   location_codes.put("uaspcoll", "University of Alberta Bruce Peel Special Collections");
   location_codes.put("uatechserv", "University of Alberta Bibliographic Services");
-  location_codes.put("vanguard", "Vanguard College");
+  location_codes.put("uarcrf", "University of Alberta Research and Collections Resource Facility");
 
   for (String code : codes){
     locations.add(location_codes.get(code.toLowerCase()));
```
pgwillia: Updated to the values in the location.yml config. Assuming that they are up to date.

config/SolrMarc/symphony_index.properties
```diff
@@ -142,7 +142,7 @@ source_tesim = "Symphony"
 holdings_tesim = 949[a-z]
 owning_library_tesim = 999m
 #location_tesim = 926m
-location_tesim = script(locations.bsh), getLocations
+location_tesim = script(location_facet.bsh), getLocationFacet
```
pgwillia: Using the beanshell script that maps the library location codes to its description.

spec/features/advanced_search_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    
+    # issue:1178 Library locations should be sorted in alpha order
+    # previously these were stored as codes (ie uaeduc, uabusiness) in the index and then 
+    # mapped to their full title (ie University of Alberta HT Coutts Education, University of Alberta Winspear Business)
+    click_link 'Library'
+    expect(page.body.index('University of Alberta HT Coutts Education')).to be < page.body.index('University of Alberta Winspear Business')
+    within '#facet-location_tesim' do
+      libraries = all('li span.facet-label').map(&:text)
+      expect(libraries.sort).to eq(libraries)
+    end
```
pgwillia: Thought of two ways to test this.  One uses two values that were out of order previously. The other solution just checks that the resulting text is sorted alphabetically. Preferences?

nnunn: I think just checking that the resulting text is sorted alphabetically makes the most sense to me

config/SolrMarc/index_scripts/location_facet.bsh
nnunn: I had to make changes to location.yml at on August 23 (pr #1237) - would these changes be reflected here?

pgwillia: No, they wouldn't be.  I'll take a look.

pgwillia:
I found some other code that refers to locations that can be simplified.

PR #1227, opened 2018-08-13
Discovery: move ingest Solr target out of config/ingest.yml

pgwillia:
* rename config/solr.yml to blacklight.yml to address deprecation
* move config to use environment variables set in secrets.yml
* add lightweight docker containers with README instructions 
* re-introduces some tests that were previously failing


pgwillia:
resolves #1229

pgwillia:
There's some stuff here that people working on ingest related tasks, writing tests for bug fixes, and working with a local solr will appreciate. BUT, there is significant changes to deployment that still need to be discussed with Neil.

pgwillia:
I've rebased on the latest master and rolled back the changes that would hinder deployment.

.travis.yml
```diff
@@ -1,9 +1,24 @@
...
+  chrome: stable
 rvm:
   - 2.1.5
+  - 2.4
 cache: bundler
 bundler_args: --without development production --jobs=3 --retry=3
 before_script:
+  # Disable services enabled by default
+  # http://docs.travis-ci.com/user/database-setup/#MySQL
+  - sudo /etc/init.d/mysql stop
```
murny: Small suggestion but you could just do what we do on jupiter and leave mysql commented out in the lightweight.yml? Feel like any developer will have mysql/postgres running on their machine anyways. 

Then you can use the database that comes with Travis (which is far better, since docker containers/images don't cache on travis yet...so this adds alot of overhead on your builds)

pgwillia: If I recall correctly I tried this first, but there was a hiccup with config for the mysql that comes with travis.

pgwillia: Must have been a faulty memory.  Now uses the db that comes with Travis!

PR #1233, opened 2018-08-17
Some last minute fixes

nnunn:
adding ejournals and databases into main search (for the bento removal). Fixing library codes in locations.yml

### DIFF
--- a/config/locations.yml
+++ b/config/locations.yml
@@ -20,7 +20,7 @@ ahssturg: Alberta Health Services - Sturgeon Community Hospital
 ahstbcc: Alberta Health Services - Tom Baker Cancer Centre
 ahsglenrse: Alberta Health Services - Glenrose Rehabilitation Hospital
 aitfcalg:  Alberta Innovates - Technology Futures - Calgary
-aitfc-fer: Alberta Innovates - Technology Futures - C-FER Technologies
+aitfcfer: Alberta Innovates - Technology Futures - C-FER Technologies
 aitfdevon: Alberta Innovates - Technology Futures - Devon
 aitfmw: Alberta Innovates - Technology Futures - Mill Woods
 aitfveg: Alberta Innovates - Technology Futures - Vegreville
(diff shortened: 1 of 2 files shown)


config/locations.yml
```diff
@@ -32,6 +32,7 @@ covenantmh: Covenant Health - Misericordia Community Hospital
 cross:  Cross Cancer Institute
 cuc:  Canadian University College
 gprc: Grande Prairie Regional College
+gprcgp: Grande Prairie Regional College
```
weiweishi: Is this intended? to have two codes for Grande Prairie Regional College?

weiweishi:
sounds good.


*Weiwei Shi*Digital Initiative Applications Librarian

2-10L Cameron Library, University of Alberta
780-492-7802  |  weiwei.shi@ualberta.ca
*"The University of Alberta respectfully acknowledges that we are situated
on Treaty 6 territory, traditional lands of First Nations and Métis
people."*

config/locations.yml
nnunn: Sadly this isn't the most recent list Sean -- there are codes that don't appear in that list that are coming up on search. For example -- ahserc doesn't appear on the list but does appear in the results. I have added a listing for it as Alberta Health Services, but this is a duplication... thoughts?

seanluyk:
If Elaine is around it might be worth asking about specific codes. She sent
the list to me just this week.

Sean Luyk

nnunn:
Ok that last commit fixes all the library codes *but* there are duplications of some libraries. Especially Alberta Health Services (I did remove the Grande Prairie duplication as I think they have just split their location into three different locations). I also had to add a really strange location which is obviously just a mistake - usic tech lab: University of Alberta Cameron - Science & Technology - Lab. This at least gets rid of the awful looking symphony codes in the library facet, but I will check with Elaine about the missing codes. I think though we should still get this version into test right away.

seanluyk:
Agreed and we'll need to work out a process for keeping these up to date.
Thanks Natasha!

Sean Luyk

PR #1235, opened 2018-08-21
Natasha minor aug 20

nnunn:


app/views/catalog/_subjects.html.erb
```diff
@@ -8,13 +8,13 @@
             <% i = 0 %>
             <% sbstring = "" %>
             <% while( i <= index ) %>
-              <% sbstring = sbstring + subject[i].gsub("&","").gsub(" ", "+") %>
+              <% sbstring = sbstring + subject[i].gsub("&","").gsub(".", "") %>
```
weiweishi: So we no longer needs to worry about the space character in the search query? if we remove this `gsub(" ", "+")`?

nnunn: The spaces do not impact the query so I don't think we have to worry about them.

Issue #938, opened 2016-04-01
Authors/additional authors - different results from advanced vs. find more by this author search


seanluyk:
Clicking on find more by this author appears to only return results where the term is in the author field, but not additional author fields. However, an advanced search for an author returns results that include the term in the additional author field.

For example, clicking on [find more by this author from the Darnton book](https://search.library.ualberta.ca/catalog?f%5Bauthor_display%5D%5B%5D=Darnton,+Robert) returns 31 results, but an advanced search for [Darnton, Robert](https://search.library.ualberta.ca/catalog?all_fields=&author=Darnton%2C+Robert&commit=Search&op=AND&page=2&range%5Bpub_date%5D%5Bbegin%5D=&range%5Bpub_date%5D%5Bend%5D=&search_field=advanced&sort=score+desc%2C+pub_date_sort+desc%2C+title_sort+asc&subject=&title=&utf8=%E2%9C%93) returns 38 results, 7 of which are for titles where this author is an additional author.

It would be ideal if these worked the same way, and I think it's useful to include titles where the author is an additional author. What does everyone on @ualbertalib/discovery-testing think?

redlibrarian:
I agree that the functionality should be consistent. I'm not sure which (Solr) fields the advanced search box looks at, but I should be able to add them to the facet for "more by this author".

theLinkResolver:
@seanluyk Agreed.

linds37:
also agree that both search functions should pull results from the author and additional author fields.

redlibrarian:
Done when:

- [ ] Discrepancy identified
- [ ] Problem solved (see also issue #886 )

weiweishi:
Should already be fixed in Release 3.0.85.

Issue #1115, opened 2017-05-15
Problems printing record views in Chrome


redlibrarian:
See: https://blog.donnierayjones.com/2014/10/fix-bootstrap3-printing/

seanluyk:
See also https://github.com/ualbertalib/NEOSDiscovery/issues/146

weiweishi:
This is fixed in Release 3.0.85.

Issue #1109, opened 2017-04-11
Subject terms not working


redlibrarian:
I am having trouble locating items on the Michif language in the discovery system.

Here is a known item I already have:

https://www.library.ualberta.ca/catalog/7120087

when I click through to see if there are more like it, using the link at the bottom in the Subjects:
 
Michif language  > Juvenile sound recordings
I am just getting a random list of Juvenile sound recordings, rather than ones that are in Michif.  

129 items and all various languages.  

The subjects search in the record are not working correctly.

seanluyk:
Re: the subject chaining issue. v3.0.82. It may not actually be fixed in test after all. If you click on [Michif language > Juvenile sound recordings](https://search-test.library.ualberta.ca/catalog?q=Michif+language+%2B+Juvenile+sound+recordings&search_field=subject&view=brief) it's acting as a keyword search within the subject field (the second result is teach me Italian). If I add [quotes to the URL](https://search-test.library.ualberta.ca/catalog?q=%22Michif+language%22+%2B+%22Juvenile+sound+recordings%22&search_field=subject&view=brief) I believe it's in line with expectations for how subject searches work. We should talk about this in person though before making any changes as this has some usability implications

seanluyk:
Here's an example of a URL constructed to do an exact match: https://searchworks.stanford.edu/catalog?q=%22English+language+Dictionaries+Japanese.%22&search_field=subject_terms

seanluyk:
Closing is this is being solved in the next release/this information is duplicated elsewhere

Issue #910, opened 2016-02-24
Clicking on subjects from database records leads to something has gone wrong page


seanluyk:
Just encountered this when clicking on the library and information science subject from this record:

https://www.library.ualberta.ca/catalog/11710953

redlibrarian:
Hm. This seems to not affacet all database records. It looks like it either affects records with multi-word subjects, or subjects with "&" in them.

redlibrarian:
Problem was with the ampersand. I've fixed this - it will be in the next release.

seanluyk:
Interesting, thanks!

redlibrarian:
This is working in test now.

seanluyk:
This is working, but I'm now wondering why clicking on a subject attached to a database brings users to the grid view? The expected behaviour would be to see more databases with that subject applied in a list view, I think

seanluyk:
As the initial bug has been fixed, this can be closed. I'll create a new issue about the usability question

PR #1248, opened 2018-09-04
…

nnunn:
fixing printing bug, adding quotes around subject searches, fixing author searches.

### DIFF
--- a/app/helpers/subject_helper.rb
+++ b/app/helpers/subject_helper.rb
@@ -1,16 +1,13 @@
 module SubjectHelper
-
-   	def subject_query(subject, index)
-   		i = 0
-        sbstring = "" 
-        while( i <= index )
-       		sbstring = sbstring + subject[i].gsub(".","").gsub("&","").gsub("+"," ")
-            	unless i == index
-                	sbstring = sbstring
-                end
-            i+= 1
-        end
-   		subject_query = "\"#{sbstring}\""
-   	end
-   	
+     def subject_query(subject, index)
+       i = 0
+       sbstring = "" 
+       while i <= index
+          sbstring += subject[i].delete(".").delete("&").tr("+", " ")
+          sbstring = sbstring unless i == index
+          i+= 1
+       end
+       sbstring = "\"#{sbstring}\""
+       sbstring
+     end  
 end
--- a/app/helpers/subject_helper.rb
+++ b/app/helpers/subject_helper.rb
@@ -1,0 +1,16 @@
+module SubjectHelper
+
+   	def subject_query(subject, index)
+   		i = 0
+        sbstring = "" 
+        while( i <= index )
+       		sbstring = sbstring + subject[i].gsub(".","").gsub("&","").gsub("+"," ")
+            	unless i == index
+                	sbstring = sbstring
+                end
+            i+= 1
+        end
+   		subject_query = "\"#{sbstring}\""
+   	end
+   	
+end
--- a/app/helpers/subject_helper.rb
+++ b/app/helpers/subject_helper.rb
@@ -1,13 +1,8 @@
 module SubjectHelper
-     def subject_query(subject, index)
-       i = 0
-       sbstring = "" 
-       while i <= index
-          sbstring += subject[i].delete(".").delete("&").tr("+", " ")
-          sbstring = sbstring unless i == index
-          i+= 1
-       end
-       sbstring = "\"#{sbstring}\""
-       sbstring
+     def subject_query(subjects)
+       # clean up characters than can mess with queries
+       subjects = subjects.map { |subject| subject.delete('.&').tr('+', ' ') }
+       # deal with subject chaining and make a phrase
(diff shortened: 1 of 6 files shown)


pgwillia:
I'm about to take a look at this.  Just wondering what was the problem and expected behaviour with the "printing bug?"  Similar questions about subject and author searches; What was the problem and expected behaviour?

nnunn:
The full record was not printing in Chrome. This was fixed by removing the bootstrap class "hidden-xs". See issue #1115 

The subject search within a full record needed to be in quotes so that the exact string was searched. see issue #1109 

The author search fixes issue #938  - the author link within a record should return the same results as an advanced search of that author.

Hope that helps - sorry I wasn't clearer in the initial pull request.

nnunn:
If you print a record view from Chrome (for example https://library.ualberta.ca/catalog/3053737) you only get the table holdings and not the metadata for that record. This only happens in chrome and it appears it is because chrome is hiding divs with the bootstrap class "hidden-xs" - since we actually don't need the record metadata hidden on small devices, I removed this class. This now prints the full record metadata in Chrome.

I can move the string manipulations into a helper - are you specifically talking about the subject string manipulation?

I don't actually know how to set up regression tests...

app/views/catalog/_subjects.html.erb
```diff
@@ -8,13 +8,13 @@
             <% i = 0 %>
             <% sbstring = "" %>
```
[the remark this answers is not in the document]
nnunn: sure that makes sense. I'll do that and push it up.

nnunn:
That would be great - thanks! Maybe after you set it up you could walk me through it so I know how to do it next time?

nnunn:
Ok, I've created and pushed the subject helper. Let me know if it looks ok.

Do you have any further questions about the printing bug?

pgwillia:
@nnunn Sorry I didn't get back to this today.  If you look at the "All checks have failed" you can see that Rubocop has some feedback on the code style in the helper.  A quick way to resolve most of those issues is to run `rubocop -a` which auto-corrects the nags when possible.

Still can't reproduce the chrome bug, but the change seems logical to me.

I'll look at writing the test tomorrow. The ability to test is improved by https://github.com/ualbertalib/discovery/pull/1227, but it there is a change to how the application is configured that would have to be matched by Neil in ansible.

app/helpers/subject_helper.rb
```diff
@@ -0,0 +1,13 @@
+module SubjectHelper
+     def subject_query(subject, index)
```
pgwillia: Great!  Thanks this makes it easier to read, test and talk about this code.

I think there are some room for improvements here.  First at `app/views/catalog/_subjects.html.erb:7` by passing `subject_query(subject.take(index+1))` we can change the method to `def subject_query(subjects)` and not have to use index or a counter to reason about the arrays of strings.

Second, we can use `subjects = subjects.map {|subject| subject.delete('.&').tr('+', ' ')}` to clean up the string.

Third, we can use `subjects.join.strip` to replace `sbstring +=`

Finally, I think making this a phrase replaces the need to do some of the string cleanup. I did some digging and the & is from [here](https://github.com/ualbertalib/discovery/issues/910), but I couldn't see where . and + were important. Do you know? Is there a good example?
If I'm right then this method becomes
```
def subject_query(subjects)
  "\"#{subjects.join.strip}\""
end
```
Here's an example of the test that I've written for this helper.
```
# spec/helpers/search_helper_spec.rb
require "spec_helper"

describe SubjectHelper do
  describe "#subject_query" do
    it "should handle ampersands, periods and plus" do
      # https://github.com/ualbertalib/discovery/issues/910
      expect(helper.subject_query(['Library & Information Studies'])).to eq "\"Library  Information Studies\""

      expect(helper.subject_query(['England and Wales. Parliament'])).to eq "\"England and Wales Parliament\""
      expect(helper.subject_query(['C++ (Computer program language)'])).to eq "\"C   (Computer program language)\""
    end
    
    # https://github.com/ualbertalib/discovery/issues/1109
    it "should handle subject chaining" do
      expect(helper.subject_query(['Michif language ', ' Juvenile sound recordings'])).to eq "\"Michif language  Juvenile sound recordings\""
    end
  end
end
```

nnunn: Great - that is a lot cleaner. Thanks!

So if I were to create a test like that how would I actually run it?

nnunn: I don't think the + and . break anything, but the query strings look a litte odd with them in there. I think from a user perspective it makes sense to strip them out.

pgwillia: I've added these changes as a commit to this branch.  If you `git pull` you should have the test.  You can run it and all the tests `bundle exec rake spec`

nnunn:
Are there any other changes I should make on this pr?

app/helpers/subject_helper.rb
nnunn: awesome thanks!

Issue #1209, opened 2018-08-08
Discovery: Deployment of release 3.0.85


pgwillia:
After the following (at a minimum) have been merged
- [x] #1200 - bug fix to display library locations in alphabetical order (requires re-index)
- [x] #1248 - which addresses #1115 (printing issue), #1109 subject search as a phrase, #938 expanding scope of author search
- [x] bump gem versions: better-errors, rubyzip, turbolinks, factory-bot, paper-clip, addressable

seanluyk:
This is very exciting, great work, everyone!

*Sean Luyk <https://sites.google.com/a/ualberta.ca/sean-luyk/> *[lʌʊ̈k
<https://en.wikipedia.org/wiki/File:Nl-Luik.ogg>]*, MA, MLIS*
Discovery & Web Services Librarian (interim)
780-492-6779

weiweishi:
Question from @nmacgreg and @seanluyk

```Re: https://github.com/ualbertalib/discovery/releases/tag/3.0.85
Can you elaborate on the "... (requires re-index) ..." feature of this new version?

It happens that today is the day of the monthly re-indexing, so I'd like to know whether I should proceed, or wait.
```

The deployment doesn't depend on a reindex. But to make the bug fix in effect it needs a reindex. So I think if the deployment happens before the monthly reindex, we will kill two birds with one stone. But otherwise, we will need to schedule another reindex to make the change take effect. @pgwillia could you please verify?

pgwillia:
>Can you elaborate on the "... (requires re-index) ..." feature of this new version?


I outsmarted myself with this.  I had left the code that maps the locations to the text in until Friday afternoon, but opted to remove it as to catch the whole issue at once.  There is now no way to deploy this update, to achieve the intended bugfix, whilst maintaining uptime.  Thanks for your caution.

How are updates to NEOSDiscovery being handled at the moment? The same bug seems like it would exist there too.

Shall we revert my bugfix and deploy the rest?


>Ugh.

Mmhm.

>1. Does this version require a change to the Solr schema?  (I'm guessing not, because there have been no updates to https://github.com/ualbertalib/blacklight_solr_conf )

 It does not.

>2. Since the NEOSDiscovery application also uses the same Solr collection, is it in any way affected ? 

[...]

PR #1262, opened 2018-09-11
fix unexpected underscores and dashes in location codes

pgwillia:
#1209


theLinkResolver:
Are you finding irregularities in the Symphony data or are the issues somewhere? (I don't expect irregularities in this particular Symphony data, so I'd like to know if that's not the case)

pgwillia:
@theLinkResolver the issue was with an assumption I made, not being familiar with the data. I used [this marc subset](https://github.com/ualbertalib/discovery/blob/master/spec/fixtures/subset.mrc) to test which didn't contain library locations that differed from the ones [in this list](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/index_scripts/location_facet.bsh#L11-L90). It wasn't until production data was reindexed that I realized my incorrect assumption. Some codes appear like 'AITF_C-FER' instead of 'aitfcfer'

Issue #1137, opened 2017-11-30
Remove Bento Box and Article Integration (epic)


redlibrarian:
**Requirements**

As a user I want to:
- [ ] Go *either* to locally indexed collections *or* to Articles.
- [ ] Want a single, unambiguous advanced search in each search area.
- [ ] Want consistent, intuitive facets in the local search.
- [ ] Want facets organized more intuitively.

As Web Librarian I want to:
- [ ] Separate Articles search from local search
- [ ] Present local search and articles search unambiguously on the web page

This requirement comes out of the Usability Study (see #1145). The study recommends removing the bento box and separating the article search from the main discovery search. The core requirements of this issue are:

- [x] Remove bento controller and views
- [ ] Remove EDS gem and helpers
- [ ] Remove EDS advanced-search controller and views
- [ ] Clean up routes to reflect absent bento and EDS functions.

On the website side, we will need to design a new link to EDS for article searching. We could also consider dropping EDS entirely, and presenting an EBSCOhost link for Article searching (CSU would have to be consulted on whether the content was comparable) - this would be down the road.

These changes are currently being considered by WAT. @nnunn is working on wireframes for approval (and can include them in this issue when approved). 

A lot of issues will be closeable when we de-integrate Article search:
#1126 
#1091 
#1066 
#697 
#442 
#425 
#1107 
#1057 
#1056 
#993 
#929 
#911 
#649 
#415 
#414 
#151
#412
#413
#577

redlibrarian:
Update: Since the release of the Discovery usability report, we have decided to remove the bento box and the articles integration, so this issue is obsolete. (Leaving open until our meeting this afternoon).

redlibrarian:
We might be able to change this issue to "remove EDS gem and integration".

weiweishi:
Article search is still in place but the UI changes on bento has been completed. Closing the ticket.

PR #1269, opened 2018-09-13
clean up eds and article dead code

### LINKED ISSUE #1228: Remove Article integration
I am separating this issue out from removing the Bento box. Once we are certain that we are no longer using any of the articles code we should remove the views, controller, and config files.

pgwillia:
closes #1228 and addresses most of #1137

I took a quick look at use of [articles in css](https://github.com/ualbertalib/discovery/blob/master/app/assets/stylesheets/ualib-discovery.scss#L480-L493) but seemed to be tied to all facet display.

Also boosts coverage to `804 / 1178 LOC (68.25%) covered.`


app/controllers/application_controller.rb
```diff
@@ -13,24 +13,11 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   def after_sign_in_path_for(resource)
```
murny: We probably could just remove both of these devise method overwrites entirely? From a quick `ctrl-f` nothing sets `@results` or `@current_url` that I can find...and the default behaviour is to go to the `root_url`

config/routes.rb
```diff
@@ -48,59 +42,4 @@
 
   # Make sure this routeset is defined last
   comfy_route :cms, :path => '/', :sitemap => false
-
-  # The priority is based upon order of creation: first created -> highest priority.
```
murny: Yes! Thanks for removing this giant comment 👍

pgwillia:
Thanks @murny. Didn't know about the .keep -- but that makes sense now that you mention it. I removed the methods you mentioned.  I agree that the initialization doesn't seem to be used.

PR #1277, opened 2018-09-21
Add link for librarian view - fix#1242

murny:
Hey @weiweishi, tried to throw a quick PR for this together. 

However I couldn't get my local discovery to get up and running for some reason (issue with solr...think docker image is broken? It's only creates the test core, not the development core :thinking:) . So wasn't actually able to test this. 

But this _should_ work or at least give you an ideal of whats needed.

Also not sure if natasha is working on this? She assigned herself to this and put the ticket in progress?

 These catalog views folder needs a heavy refactor....we shouldn't have 70 views and partials for this! :astonished: https://github.com/ualbertalib/discovery/tree/master/app/views/catalog. Majority of these are 1-3 line partials that are never used more than once :disappointed:

### DIFF
--- a/app/views/catalog/_show_tools.html.erb
+++ b/app/views/catalog/_show_tools.html.erb
@@ -27,14 +27,17 @@
         <%= render "endnote" %>
       <% end %>
 
+      <li>
+        <%= link_to 'Librarian View', librarian_view_catalog_path(id: @document), target: :_blank %>
+      </li>
 
-        <%- if @document.respond_to?( :to_email_text ) -%>
-          <%= render "email" %>
-        <%- end -%>
+      <%- if @document.respond_to?( :to_email_text ) -%>
+        <%= render "email" %>
+      <%- end -%>
 
-        <%- if @document.respond_to?( :to_sms_text ) -%>
-          <%= render "sms" %>
-        <%- end -%>
+      <%- if @document.respond_to?( :to_sms_text ) -%>
+        <%= render "sms" %>
+      <%- end -%>
 
 
       <li class="send_correction">


weiweishi:
LGTM. 
screen shot 2018-09-21 at 7 23 25 pm

screen shot 2018-09-21 at 7 26 01 pm

@seanluyk Right now it opens in a new tab instead of a pop-up window, which may require more work. Please let me know if this works for you. It is available on the UAT server.

weiweishi:
I think @nnunn was going to work on this but had to work on other priorities. So Sean and I decided if this can be a quick fix to fit into this release, we will try to have it done before the release is cut. I will open a ticket for the refactoring of views - thanks very much for your work on this.

app/views/catalog/_show_tools.html.erb
```diff
@@ -27,14 +27,17 @@
         <%= render "endnote" %>
       <% end %>
 
+      <li>
+        <%= link_to 'Librarian View', librarian_view_catalog_path(id: @document), target: :_blank %>
```
weiweishi: Ideally, this should be i18n? but there are lots strings in these partials that are not i18n-ed so it would be a big separate refactoring work

murny: Yep I agree. In Discovery we only using i18n for things that are using the same definitions  from the blacklight gems. So if you look in the locale folder (https://github.com/ualbertalib/discovery/blob/master/config/locales/) there's nothing here or for example the en.yml (https://github.com/ualbertalib/discovery/blob/master/config/locales/en.yml)

Probably could do majority of this when we do the refactor of the views for catalogs

PR #1279, opened 2018-09-22
correct the capitalization for the view citation label - fixes #1058

### LINKED ISSUE #1058: lowercase letters in "MLA" and "APA"
@ualbertalib/discovery-testing 

Change to captalized "MLA"/"APA"

apamla

weiweishi:
The translation/i18n in blacklight in missing completed. Don't know why this one uses i18n string. This is not the right way to fix the issue, but it is the quickest way. And we will need to deal with i18n issue more holistically when/if we refactor the views.


weiweishi:
thank you so very much @murny . Totally not expecting you to work on a Friday night!!!

PR #1300, opened 2018-10-01
remove held_by in favour of locations

pgwillia:
https://github.com/ualbertalib/NEOSDiscovery/issues/152
screenshot from 2018-10-01 12-57-16


app/controllers/application_controller.rb
```diff
@@ -17,7 +17,6 @@ def load_lookup_tables
     @languages ||= YAML.load_file("#{Rails.root}/config/languages.yml")
     @statuses ||= YAML.load_file("#{Rails.root}/config/statuses.yml")
     @item_types ||= YAML.load_file("#{Rails.root}/config/item_types.yml")
-    @library_codes ||= YAML.load_file("#{Rails.root}/config/library_codes.yml")
```
weiweishi: So this file was used only for this one purpose? to supply library code for "copied owned/held_by"? Wow!

app/views/catalog/_index_default.html.erb
```diff
@@ -20,14 +20,8 @@
...
         <dd class="blacklight-format">
-          <% if document["held_by_tesim"] %>
-            <% @held_by_locations = [] %>
-            <% document["held_by_tesim"].first.split.each do |held_by| %>
-              <% @ual = true if (@library_codes[held_by.to_i].include?("University of Alberta") or (@library_codes[held_by.to_i].include?("NEOS Free"))) %>
-              <% @held_by_locations << @library_codes[held_by.to_i] %>
-            <% end %>
-            <%= @held_by_locations.join(", ") %>
-          <% end %>
+          <%= document["location_tesim"].join(', ') %>
```
weiweishi: Seems that there were some exceptions like "University of Alberta" or "NEOS Free". Are those presented in the location_tesim? Do we need to treat them separately?

pgwillia: Thanks for the reminder.  I didn't follow up on moving the @ual shield.  Will move this to a helper and add some tests.

pgwillia: Fixed this!

app/controllers/application_controller.rb
pgwillia: As far as I can tell...  Doesn't appear anywhere else, and searching the history on github doesn't make it super clear how or why this was added.

Issue #1260, opened 2018-09-10
Implement RCRF No Loan Form


weiweishi:
Once the wireframe is finalized here, #1216, the form needs to be developed on the front end:

https://balsamiq.cloud/s3awdua/pmnl7qx/r8C53

**Requirements**
Auto-populated fields:

- the title of the object
- call number of the object.

The request form should be sent to depository@ualberta.ca, and ill@ualberta.ca

Also question for the URL for this form? Could we do something like:
https://library.ualberta.ca/rcrf-noloan-form or https://library.ualberta.ca/rcrf-noloan-request ? etc?

seanluyk:
@weiweishi my preference is /rcrf-readonsite. I know the form is labeled no loan, but we'll need to talk to CJ about this. Since the location/status is called ReadOnsite, I think it'll be confusing for users if we use the term no loan to refer to these materials

murny:
One thing we need to think about (for all three of these request form tickets: #1260, #1259 and #1258) is how emails are sent in our different environments. Often times a bad application will send real emails to real people on UAT/Staging environments! Which is something we should avoid doing. So for the different environments here's what we need to do to prevent this:

#### Development
By default development should not be sending emails. Typically in a rails application, the email is outputted to the logs. You can bring something like `letter_opener` (https://github.com/ryanb/letter_opener) which allows the email to be nicely previewed in a new tab in the browser instead.

#### Test
Typically you don't have to worry about test environment. We should be configuring this using `ActionMailer::Base.delivery_method = :test` which makes emails not send out in test environment but instead puts the email on an array instead. More info here: https://guides.rubyonrails.org/testing.html#testing-your-mailers

#### UAT/Staging
We should write a [email interceptor](https://guides.rubyonrails.org/action_mailer_basics.html#intercepting-emails), which intercepts any email that is being send out by discovery, clears the `to` fields and sends the email to `dittest@ualberta.ca` (or something else?) instead. You can also append some comments in the body of the email explanaing why this happened (this was sent out in a non production environment) and who this email was originally for, etc.

#### Production
[...]

Issue #1184, opened 2018-06-25
Discovery: BPSC Request Form


weiweishi:
**High-level user story**
As a user of the Discovery service, I want to be able to send a request to the Bruce Peel Special Collections to view non-requestable/non-loanable materials held there. It needs to allow me to:
- Select where I want to view the material (RCRF or BPSC)
- Notify BPSC staff with my contact information so that they can contact me and book an appointment with me for me to come in and view the material
- Autopopulate bib-record information into the form.

**Current situation**
Currently, when the user sees a non-loanable/non-requestable item, they can't take any action on the item to proceed with a request. 
![image](https://user-images.githubusercontent.com/2047614/41866353-1bcf595a-786d-11e8-90aa-84f7fd44c5a8.png)

There are two types of materials that can be requested at BPSC:
- Monographs/Books 
- Archival materials (currently about 12 archival collections, but numbers will be actively increasing as they have staff working in this area)

BPSC currently has two forms, one for requesting monographs/books: https://docs.google.com/forms/d/126ZJ6r6L42AeoWVyOQRWdY9b0SYv7VxUljx6V1d56Po/viewform?edit_requested=true
one for requesting archival materials:  https://docs.google.com/forms/d/1-L6cAek91koI7lU5_4j6kpp_4-AUYIhETM26Xjg1uEw/viewform?edit_requested=true

Archival collections have the finding aid link on their MARC record, that differentiating them from regular BPSC objects. 
https://neos.library.ualberta.ca/uhtbin/cgisirsi/?ps=RkqsM66Ulk/UAARCHIVES/67260015/123

This information should be part of the API response from Symphony, but it is not presented on the Discovery record. 

The list of current available archival collections is on the Discover Archives Retrieval Request Form (Bruce Peel Special Collections) form. 

Both forms are Google Forms, which gives the BPSC staff a spreadsheet to help them track the requests. 

**Detailed requirements**

- Add the "Request Me" button on an object that has a "ReadOnSite" status, and belongs to Bruce Peel Special Collection. 
![screenshot from 2018-06-25 11-52-14](https://user-images.githubusercontent.com/2047614/41867013-1eced8f4-786f-11e8-971b-a11ded7d192d.png)
- Required fields on the form:
  - Name/Email of the person who sends the request
  - appointment time for the viewing
- For monograph/book materials (the majority of the requests), the form should be populated with title and call number of the object. An additional optional field should be a free-text field for other information so that the form can be used outside the discovery workflow. 
[...]

Issue #1297, opened 2018-09-27
[Discovery] Use database to manage library location codes/status codes


weiweishi:
Currently, library location codes and status codes are managed in multiple yaml files, and each time when they change (and they change often!), it requires a new release of the Discovery service. We need to find a better way to manage these library codes/status codes. 

Proposal: Create a table in the application database, create models for library status codes and location codes, and use activerecord to access these values? Any update will be a change to the DB, instead of deployment of code. 

 Questions:
- do changes to the databases need to be loaded immediately by the application? Or how often?
- who changes the database and how? Is an admin interface needed?
- is it possible to pull the information directly from Symphony, with no intermediate database? (we already have a connection to Symphony web services)

seanluyk:
Out of scope pending discovery strategy recommendations

pgwillia:
1273

pgwillia:
Some thoughts (definitely not a stupid question) from @nmacgreg 

>Stupid question: can't you just get the library + status codes you need, directly from an API in Symphony? (guessing that the data gets loaded there, first).  Or make it part of the nightly "incremental ingest" process?  (Let me guess: not enough source fields)

>The shared information, between NEOSDiscovery & Discovery, is tricky.  Logically, these apps should share those tables, but they already have 1 database each.  Does that mean we should create a 3rd database, which they share?  I dimly recall that Rails supported the idea of talking to 2 separate databases... dunno if that messes up your existing code, having to specify which database, rather than the database?  Anyway, this is out of my domain, I'll shut up now.

pgwillia:
@redlibrarian says
>The library and status codes *are* maintained in Symphony, but my understanding of the search API is that it's not a full API against all the data Symphony contains. Rather, it's a search API for items; when I was going through the API documentation to build the API calls Discovery is currently using, those were the only search functions I could find. However, I know there are other APIs for things like patron records, transactions, etc, so it's possible that that configuration information *is* exposed somewhere and I just never saw it. On the other hand, it's also possible that SirsiDynix never thought anyone would want to query the full list of codes and so haven't exposed them via an API.
> If they aren't available over the API, they may be exportable - I'd have to talk to the ILS team to find out.

pgwillia:
- [ ] SPIKE - can you have the same migration in two applications
- [ ] SPIKE - gemify?
- [ ] Implementation of NEOS Discovery
- [ ] Refactor for SolrMarc

PR #1305, opened 2018-10-03
request me button to link to readonsite forms

### LINKED ISSUE #1247: [Discovery] Business logic to identify objects that need to have No Loan form
Based on the requirements, any object that is located at RCRF, and has a location/status "READONSITE" needs to have the RCRF no-loan form displayed #1251. Any object that is located at BPSC and has a location/status "READONSITE" needs to have the BPSC no loan form displayed #1184 

 

The information needs to be either parsed from the API response
(BPSC no-loan item: https://ws.library.ualberta.ca/symws3/rest/standard/lookupTitleInfo?clientID=Primo&marcEntryFilter=ALL&includeItemInfo=true&includeAvailabilityInfo=true&includeMarcHoldings&titleID=5642450
RCRF no-loan item: https://ws.library.ualberta.ca/symws3/rest/standard/lookupTitleInfo?clientID=Primo&marcEntryFilter=ALL&includeItemInfo=true&includeAvailabilityInfo=true&includeMarcHoldings&titleID=8081485)

Note that "current location" (not home location) is what needs to be used to identify these items, in addition to libraryID.

Or we could obtain that information from Solr, in the 926 tag, which is currently indexed but not displayed: https://library.ualberta.ca/catalog/8081485/librarian_view. Catch is the data in 926 is not live, and has a day's delay. 

Blocked by the implementation of the READONSITE status code in Symphony.

pgwillia:
closes #1247

screenshot from 2018-10-03 14-30-26
screenshot from 2018-10-03 14-29-12
screenshot from 2018-10-03 14-41-29

depends on #1184 and #1260 
TODO:
- [ ] replace TODO paths with appropriate `path_for`
- [ ] ensure parameters to populate the forms are passed

### DIFF
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -25,23 +25,35 @@ module HoldingsHelper
     end
   end
 
+  # TODO: If the way that libraries are mapped changes (replacing config/location.yml) this should follow that scheme
+  UAL_SHIELD_LIBRARIES = [
+    LIBRARY_LOCATIONS[:uainternet],
+    LIBRARY_LOCATIONS[:neosfree]
+  ].freeze
+
+  # determines whether or not ual shield should be displayed for the item
   def display_ual_shield(document)
     return false unless document["location_tesim"]
-    return true if document["location_tesim"].include?(I18n.t('ual_shield_UAL'))
-    return true if document["location_tesim"].include?(I18n.t('ual_shield_NEOS'))
+    UAL_SHIELD_LIBRARIES.each { |library| return true if document["location_tesim"].include? library }
     false
   end
 
   # TODO: If the way that libraries are identified changes (replacing config/location.yml) this should follow that scheme
   READ_ON_SITE_LOCATION_RCRF = 'UARCRF'.freeze # Research and Collections Resource Facility
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
-  
+
   # returns path to read on site form
   # TODO: replace placeholder with appropriate path_for
   def read_on_site_path(item)
     return '/#TODO_RCRF' if item[:location] == READ_ON_SITE_LOCATION_RCRF
     return '/#TODO_BPSC' if item[:location] == READ_ON_SITE_LOCATION_BPSC
     errors_unprocessable_path
+  end
+
+  # returns the library description that matches the code coming from symphony ws
+  # TODO: If the way that libraries are mapped changes (replacing config/location.yml) this should follow that scheme
+  def library_location(library_code)
+    LIBRARY_LOCATIONS[library_code.downcase.gsub("_","").to_sym]
   end
 
 end
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -32,4 +32,16 @@ module HoldingsHelper
     false
   end
 
+  # TODO: If the way that libraries are identified changes (replacing config/location.yml) this should follow that scheme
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -20,4 +20,5 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
```
weiweishi: Seems that we've added a new place to hold strings for location? Can we reuse any of the existing ones? If not, let's track this to make sure it will be considered as part of https://github.com/ualbertalib/discovery/issues/1297

pgwillia: Coming back to this with perspective I don't think this was an appropriate use of locales. I want to say this should have been a constant. The other places we have for locations map this code to a human readable string. This needs to identify the library. If the way we identify libraries changes with #1297, then this should be changed too.

pgwillia:
This PR uses Symphony WS rather than parsing the Marc in Solr. Should I spend more time investigating that solution?

pgwillia:
I added two commits to this PR which refactor other pieces related to the criticisms about symphony lookups.  Maybe they don't belong here but I think they improve readability and performance.

app/helpers/holdings_helper.rb
```diff
@@ -25,11 +25,35 @@ def fetch_sfx_holdings(document)
...
+  # TODO: If the way that libraries are identified changes (replacing config/location.yml) this should follow that scheme
+  READ_ON_SITE_LOCATION_RCRF = 'UARCRF'.freeze # Research and Collections Resource Facility
+  READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
+  
+  # returns path to read on site form
+  # TODO: replace placeholder with appropriate path_for
+  def read_on_site_path(item)
+    return '/#TODO_RCRF' if item[:location] == READ_ON_SITE_LOCATION_RCRF
+    return '/#TODO_BPSC' if item[:location] == READ_ON_SITE_LOCATION_BPSC
+    errors_unprocessable_path
```
murny: It would be nice if you could never reach the `errors_unprocessable_path`? If things go bad, why give the user a link to a 422 page?

You could move logic into the view maybe and avoid this entirely?

```
<% if item[:status] == 'READONSITE' %>
  <% if item[:location] == READ_ON_SITE_LOCATION_RCRF %>
      <%= link_to "Request Me",'/#TODO_RCRF', {:id => 'requestLink', :class => 'lightboxLink'} %>
  <% elsif item[:location] == READ_ON_SITE_LOCATION_BPSC %>
     <%= link_to "Request Me", '/#TODO_BPSC', {:id => 'requestLink', :class => 'lightboxLink'} %>
  <% end%>
<%- end -%>
```

Not pretty but gets the job done without any weird side effects like linking to a 422 page (it just doesn't show a link if its not one of the locations)

config/initializers/symphony_lookup_tables.rb
```diff
@@ -0,0 +1,12 @@
+# Set up the app-wide +LIBRARY_LOCATIONS+ hash from <tt>config/locations.yml</tt>
```
murny: Random question, but in the future would it be a good idea for these to be retrieved from the database? (with a CRUD admin screen so staff can edit/update/etc it)

pgwillia: Yup, https://github.com/ualbertalib/discovery/issues/1297

app/helpers/holdings_helper.rb
pgwillia: I wasn't sure what a good default would be. Perhaps it should be
```suggestion
    nil
```
Then the view can be
```
                    <%= link_to_unless "Request Me", read_on_site_path(item), {:id => 'requestLink', :class => 'lightboxLink'} %>
```

PR #1316, opened 2018-10-04
add regression test to check number of library locations

pgwillia:


pgwillia:
Right now it is 54. Sean said

> In terms of the list of libraries, it's possible to verify in elibrary:
> 
> -elibrary lists 56 libraries: one is a dead location with no items attached (Math Library), and one is just a pickup location with no items attached- Vanguard - Taylor
> -Search-test lists 54
>
> The list of libraries are correct in terms of numbers. The labels aren't all correct but that's something we can tackled in a future release!

What the test does is look at the list under "Library" on the advanced search page and counts the nodes that the xpath expression matches. I'll admit that this is a pretty brittle test.  If the layout of that page changes or the number of libraries change this test will fail. Suggestions?

nnunn:
I think it's better than nothing for now... this advanced search list is also being called from the yaml file right? Would it make sense to check that the page is displaying the number of libraries in locations.yaml?

pgwillia:
The advanced search list is from the Solr data.  I don't quite know how it's working, but that's definitely where it's coming from.

nnunn:
Oh weird -- gahh these locations are so strange. Well this will work for now -- once we've looked at how to rework the locations the tests will have to change...

PR #1319, opened 2018-10-05
removing slow code on profile page

nnunn:
Removing code which is slowing down the profile page. This code went and checked if a staff member's photo exists on www2 and if it doesn't exist displays a generic image. However, this process is slowing down the page load considerable. Therefor, I have removed it and instead added a thumbnail for all staff - generic ones for those without images.


nnunn:
Hmm I don't see that haml error.

pgwillia:
I pushed a commit for the error I was seeing.  Could you check and make sure I didn't do something silly?

PR #1337, opened 2018-10-18
Add letter opener gem to make testing emails easier

murny:
First step of many for doing the request forms:

Brings in this gem: https://github.com/ryanb/letter_opener

Email in development environment now opens a nice new tab displaying the email that was sent out:
image

Instead of hiding the contents of the email inside the console.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -37,9 +37,8 @@ Rails.application.configure do
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
   #
-  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = true
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.default_options = {from: 'no-reply@ualberta.ca'}
 end
-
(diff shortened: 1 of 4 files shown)


config/environments/development.rb
```diff
@@ -37,9 +37,8 @@
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
   #
-  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = true
   config.action_mailer.raise_delivery_errors = true
```
pgwillia: The deliveries lines in this config probably don't make sense?
```suggestion

```

config/environments/development.rb
```diff
@@ -37,9 +37,8 @@
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
   #
-  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = true
```
pgwillia: The deliveries lines in this config probably don't make sense?
```suggestion
```

config/environments/development.rb
murny: Should probably be set to false, think this is the default setting from a new rails project. Can make the change.

config/environments/development.rb
murny: This is valid, we do want to send out email in development environment (it just sends out to letter opener gem)

PR #1338, opened 2018-10-18
Add editorconfig dot file so everyone has consistent coding styles

murny:
Noticed from last PR there was a bunch of space diffs among other things. Lets align our editors to be on the same page.

This brings in an EditorConfig which helps developers maintain consistent coding styles between different editors/IDEs.

More info and links to plugins (if your editor doesnt support editorconfig) here: https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


pgwillia:
Should we correct all the code at once, or let future commits bring the current code base to a state of equilibrium?

murny:
Can do it as a separate commit. I'll look into it, from quick look, looks like there is linters that will use your editorconfig file and fix automagically your files.

PR #1344, opened 2018-10-22
Email interceptor to prevent staging/uat environments to send real email out

murny:
Preliminary work for request forms.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,46 @@ AllCops:
     - 'vendor/**/*'
   ExtraDetails: true
   TargetRubyVersion: 2.1.5
+
+# TODO: Need to bring this inline with jupiter and other projects rubocop configs.
+# Also need to fix and remove rubocop_todo.yml
+Metrics/AbcSize:
+  Enabled: false
+
+Metrics/BlockLength:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Enabled: false
+
+# Metrics/LineLength:
+#   Max: 120
+
+Metrics/MethodLength:
+  Enabled: false
+
+Metrics/PerceivedComplexity:
+  Enabled: false
+
+# Rails:
+#   Enabled: true
+
+Style/ClassAndModuleChildren:
+  Enabled: false
+
+Style/Documentation:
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -14,3 +14,46 @@ AllCops:
     - 'vendor/**/*'
   ExtraDetails: true
   TargetRubyVersion: 2.1.5
+
+# TODO: Need to bring this inline with jupiter and other projects rubocop configs.
```
pgwillia: How did the .rubocop.yml in other projects evolve?  Did you guys sit down and talk it out in advance, or just react as things came up?

.rubocop.yml
```diff
@@ -14,3 +14,46 @@ AllCops:
...
+Metrics/AbcSize:
+  Enabled: false
+
+Metrics/BlockLength:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Enabled: false
+
+# Metrics/LineLength:
```
pgwillia: Curious.  Why commented out?

murny: Cause its smaller then the one in the rubocop_todo.yml...so it will cause tons and tons of errors.

The commented out ones are left to be done later once we cleanup rubocop_todo.yml

config/secrets.yml
```diff
@@ -40,10 +40,14 @@ test:
 staging:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   solr_url: <%= ENV['SOLR_URL'] %>
+  intercept_mail: true
```
pgwillia: Does this flag add anything?  Perhaps the presence of the `intercept_mail_address` is enough.

.rubocop.yml
murny: Pretty much evolved depending on our styles. 

This one I checked in here is pretty much bare bones, which turning off some of the insane rubcop defaults (like the metric cops, documentation at the top of every file, etc), or good defaults like line length of 120, etc.

Most of the ones in jupiter project that are not in this one are ones that have evolved from bike shedding from PRs etc. We can take 5 minutes to chat about them if your curious.

Issue #934, opened 2016-03-29
Create libguide search widget


redlibrarian:
https://mail.google.com/mail/u/0/?shva=1#inbox/15380c375fd18535

redlibrarian:
Basic search widget is here: https://github.com/ualbertalib/discovery/blob/master/public/widget.html

I'll add this to the assets in LibGuides.

redlibrarian:
This has been added to libguides and can be seen on the computing science guide.  http://guides.library.ualberta.ca/computing-science

PR #1345, opened 2018-10-23
Some cleanup for discovery

murny:
- Remove files that aren't being used
- cleanup some unused gems


Gemfile
```diff
@@ -16,25 +16,19 @@ gem 'therubyracer', platforms: :ruby
...
-gem "jettywrapper", "~> 2.0.3"
 
 # Authentication
 gem "devise"
 gem "devise-guests"
 
 # Misc Utilities
-gem "addressable", "2.5.2"
 gem 'friendly_id'
-gem "htmlentities"
```
murny: These gems are never used 
- addressable already comes in via a lot of gems like rails
- htmlentities is never used
- jettywrapper is only for development/test (should not be production), and cannot be used with modern solr (5+) and is for both solr and fedora among a bunch of other things, ultimately we don't need it if we are using docker for solr for dev/test. more info here: https://github.com/samvera-deprecated/jettywrapper

.bundle/config
```diff
@@ -1 +0,0 @@
----
```
pgwillia: I agree that this shouldn't be there.  But it's initial commit gives me some pause.  I'm guessing there was some initialization/Gemfile issue at some point, probably with the development/test gems in staging when it was test and/or production.  This may resurface that error and we'll solve it differently.

murny: It did have stuff in it at one point in time, but it was removed.

Now its just an empty file. Either way this shouldn't be in the git repo. If you have custom bundler config then put it in your personal dot file or do it properly with ansible/etc

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: "from@example.com"
+  default from: "noreply@library.ualberta.ca"
```
pgwillia: Nice!

lib/widget/widget.html
```diff
@@ -1,46 +0,0 @@
-  <link rel="shortcut icon" type="image/x-icon"
```
pgwillia: Looks like this was left in as a reference? https://github.com/ualbertalib/discovery/issues/934

@nnunn got rid of [another version](https://github.com/ualbertalib/discovery/commit/9c479f5c5eb4b7e3d0dced8518b7047124e7ff7a#diff-4c9184f37cff01bcdc32dc486ec36961) but this code is still in the wild somewhere https://guides.library.ualberta.ca/computing-science.  I'll create a help desk ticket, to help us track down who's responsible for that.

murny: This is currently never being used. And because of where it is, it will never get served or bubble up to be viewable anywhere in the application. The other version you linked would get served! (as its a static file in the public directory). 

But ultimately, if this left here as a reference (which it probably is, as the code on the guides website is different), we need to do better. This git repo shouldn't be used as a dumping ground for random files. There is far better ways to store these files then the discovery repo.

pgwillia: Sorry.  I should have been more clear. I agree that it doesn't belong here. Given that someone put it there, I felt an obligation to figure out why.

PR #1353, opened 2018-10-25
Refactoring of corrections code to be somewhat sane

murny:
Much improvements to the way we do corrections for catalog items.

- cleanup routes heavily and controller actions
- fix mailing code
- UI refactor (proper bootstrap, flash messages, error handling, etc)
- Form now has proper validation
- clean up class variable usage


spec/rails_helper.rb
```diff
@@ -64,7 +64,7 @@
   #     end
   #
 
-  config.include Devise::TestHelpers, type: :controller
```
murny: TestHelpers is deprecated. Was resulting in a lot of spam when running tests.

config/routes.rb
```diff
@@ -33,8 +30,9 @@
 
   resources :staff, :as => :profiles, :controller => :profiles
 
-  resources :forms
```
murny: We were exposing every route here? when really we only needed the "show" page

app/controllers/forms_controller.rb
```diff
@@ -1,15 +0,0 @@
-class FormsController < ApplicationController
-
-  def index
```
murny: There's no corresponding view for this? So any request that goes to `/forms` 500s

config/routes.rb
```diff
@@ -33,8 +30,9 @@
...
-  post "forms/send_email" => "forms#send_email"
+  namespace :catalog do
```
murny: It be really nice to not have the item_id as a query param? But since this version of blacklight makes this difficult not sure what else we can do?
AKA:
`/catalog/1234567/corrections/new`
vs
`/catalog/corrections/new?item_id=1234567`

app/views/catalog/_correction_form.html.erb
```diff
@@ -1,2 +0,0 @@
-
-        <%= link_to "Send Correction", "/forms/catalogue_corrections?item_id=#{@document['id']}" %>
```
murny: Really need to learn how to use Rails routes helpers in discovery...we got these hardcoded routes everywhere.

app/mailers/corrections_mailer.rb
```diff
@@ -1,6 +1,6 @@
 class CorrectionsMailer < ApplicationMailer
-  def corrections_email(item_id, message, url)
-    @item_id = "#{url}/catalog/#{item_id}"
```
murny: Another hardcoded route...this isn't needed

app/views/correction_mailer/notify.html.erb
```diff
@@ -0,0 +1,4 @@
+<p>Catalog Item ID: <%= @correction.catalog_id %></p>
```
murny: This email probably needs some love with regards to the copy/text content.

Why give the ID? Wouldn't a link to the Item in discovery be better? (1234567 vs `library.ualberta/catalog/1234567`)

Maybe some context about what this is? etc?

app/views/corrections/new.html.erb
```diff
@@ -0,0 +1,35 @@
+<h1>Report Correction for Error, Typo, or Suggestion</h1>
```
murny: We should probably bring in simple_form gem, will probably do this for the other request forms

config/routes.rb
murny: Found a nice way to do this. So nevermind this! This is fixed

app/views/corrections/new.html.erb
```diff
@@ -0,0 +1,35 @@
...
+                    placeholder: 'Enter any details on the error, typo or suggestion here...',
+                    class: 'form-control' %>
+  </div>
+
+  <%= f.hidden_field :catalog_id, value: @correction.catalog_id %>
+
+  <%= f.submit 'Submit Error, Typo or Suggestion', class: 'btn btn-lg btn-default' %>
+<% end %>
+
+<p class="top-margin">
```
murny: Since we on BS3, we got no spacing utilities so just grabbed one from the discovery custom stylesheets

app/controllers/corrections_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  def new
+    @correction = Correction.new(catalog_id: params[:catalog_id])
+  end
+
+  def create
+    @correction = Correction.new(correction_params)
+
+    if @correction.valid?
+      CorrectionMailer.notify(@correction).deliver_now
+      flash[:success] = 'Thank you! We have received your correction and will be reviewing it soon.'
```
pgwillia: I18n isn't used elsewhere so feel silly trying to enforce that here.

seanluyk:
@pgwillia @murny minor UI changes look great! Also very much looking forward to this working consistently!

app/views/correction_mailer/notify.html.erb
seanluyk: @weiweishi @murny  I think all that's needed is a text box to explain the correction and a link to the item, so I don't think any additional work is needed. Here's an example of what currently comes through in the helpdesk. The subject line is what gets it put in the appropriate queue. Links do come through, not just ids:

Subject: Cataloguing Correction Submitted
Item ID: http://library.ualberta.ca/catalog/817700
Message: Bechert's name as editor doesn't show up in search results from a title search.

murny: Oops, noticed I lost the `url` to the catalog item on this email. This was due to old code calling the url...`item_id` everywhere. So left it as the ID. But can change back to giving the URL instead of an ID and fix naming.

Will fix this before merging.

config/environments/development.rb
```diff
@@ -1,4 +1,4 @@
-Rails.application.routes.default_url_options = { host: 'localhost' }
+Rails.application.routes.default_url_options = { host: 'localhost', port: 3000 }
```
murny: You want the port, so links work properly in emails and else where (this all assuming you running rails on the default port of 3000).

config/environments/uat.rb
```diff
@@ -1,3 +1,5 @@
+Rails.application.routes.default_url_options = { host: 'uat.library.ualberta.ca', port: 3002 }
```
murny: Didn't have this set for UAT. Added this in.

Issue #1282, opened 2018-09-24
Solr Docker Image only creates test core, needs dev core created as well


murny:
Using the lightweight docker-compose.yml, I am having trouble running discovery on local as Solr isn't being set up properly? It seems to only create the `discovery-test` core and not both test and dev (`discovery`) cores.

![image](https://user-images.githubusercontent.com/1930474/45966980-d5336300-bfe9-11e8-9da7-1b1fb0279337.png)

Apparently this is due to the entrypoint command wiping out the `CMD` in the dockerfile (instead of appending it). Which is deemed `won't fix` in docker-compose. More info here: https://github.com/docker/compose/issues/3140

So because of this I'd assume only `docker-compose.lightweight.yml` is affected because it uses the entrypoint command: 

```
  solr:
    image: ualbertalib/blacklight_solr_conf
    ports:
       - "8983:8983"
    volumes:
      - solr:/opt/solr/server/solr/mycores
    entrypoint:
      - solr-precreate
      - discovery-test
      - /myconfig
```

 But for the regular `docker-compose.yml` I doesn't look like a test core gets created?

```
  solr:
    restart: always
    image: ualbertalib/blacklight_solr_conf
    volumes:
      - solr:/opt/solr/server/solr/mycores
```
So this is also a problem as well, as we need both dev and test cores setup for development.

Was there a particular reason why we avoided duplicating the way Jupiter sets up the Solr docker image for development and test?

pgwillia:
@piyapongch is also experiencing this.

pgwillia:
Even after a pull, I'm not?
![screenshot from 2018-09-24 14-34-15](https://user-images.githubusercontent.com/1220762/45977479-053d2f00-c007-11e8-8dc5-e89a4ccef5b1.png)

I'm sure there's a reason why.  I couldn't find where I documented it, so it couldn't have been a good one. My best guess is that with the solr config living in a different directory (blacklight_solr_conf) that I thought it was more appropriate to create a default collection with the Dockerfile, then it would be simplest to just add another collection for the test environment with the docker-compose.lightweight.yml assuming that the default collection persisted.

My assumption has been proven false.  I support duplicating Jupiter's use of a script at the entrypoint to set up the two collections needed for development and test.

The docker-compose.yml is for the uat use case where only one Solr collection is required.

PR #1357, opened 2018-10-29
Bring over script to provision test/dev solr cores from jupiter to fix docker issue

murny:
#1282 

Temporary fix to get this working without the pain of constantly having to take down and change docker configs to run discovery in both dev and test modes.

image

But we should maybe spend some time thinking about refactoring the docker stuff to be more in align with our other apps like jupiter (or if theres improvements/better ways to do this for jupiter codebase then by all means we can improve this as well)

### DIFF
--- a/lib/docker/solr-precreate-discovery.sh
+++ b/lib/docker/solr-precreate-discovery.sh
@@ -20,20 +20,20 @@ else
     coresdir=$SOLR_HOME
 fi
 
-coredir_dev="$coresdir/development"
+coredir_dev="$coresdir/discovery"
 if [[ ! -d $coredir_dev ]]; then
-    cp -r /config/ $coredir_dev
+    cp -r /myconfig/ $coredir_dev
     touch "$coredir_dev/core.properties"
-    echo "created development core"
+    echo "created discovery core"
 else
-    echo "core development already exists"
+    echo "core discovery already exists"
 fi
 
-coredir_test="$coresdir/test"
+coredir_test="$coresdir/discovery-test"
 if [[ ! -d $coredir_test ]]; then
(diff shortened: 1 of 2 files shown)


lib/docker/solr-precreate-discovery.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
murny: Not sure if theres a better place to put this script? Put it under `lib/docker` for now.

lib/docker/solr-precreate-discovery.sh
```diff
@@ -20,20 +20,20 @@ else
     coresdir=$SOLR_HOME
 fi
 
-coredir_dev="$coresdir/development"
+coredir_dev="$coresdir/discovery"
 if [[ ! -d $coredir_dev ]]; then
-    cp -r /config/ $coredir_dev
+    cp -r /myconfig/ $coredir_dev
```
murny: I'd probably prefer `development/test` vs `discovery/discovery-test` and also `config` vs `myconfig`

But oh wells. This is needed to get this working

pgwillia: We can coordinate that change with https://github.com/ualbertalib/blacklight_solr_conf/blob/master/Dockerfile today and have config instead of myconfig.  Maybe there's something else preventing the development/test vs discovery/discovery-test, but I think to follow through the change is in secrets.yml.

lib/docker/solr-precreate-discovery.sh
pgwillia: `lib/docker` makes sense to me.  Could be `solr/scripts` to match jupiter, but I think that docker is more relevant.

PR #1359, opened 2018-10-29
WIP: Rubocop fixes attempt

murny:
TODO


app/controllers/profiles_controller.rb
```diff
@@ -1,88 +1,111 @@
...
-                :its => "Information Technology Services", :admin => "Learning Services/Libraries Administration", :science => "Science & Technology Library", :special => "Special Collections",
-                :press => "University of Alberta Press"}
-    $buildings = {:augustana => "Augustana Campus Library", :bsj => "Bibliothèque Saint-Jean", :bpsc=> "Bruce Peel Special Collections",
-            :cameron => "Cameron Library", :coutts => "Herbert T. Coutts Library ", :law => "J.A. Weir Law Library", :scott => "J.W. Scott Library",
-            :rcrf => "Research & Collections Resource Facility", :press => "Ring House 2", :rutherford => "Rutherford",
-            :stjosephs => "St. Joseph's Library", :winspear => "Winspear Library" }
-
-        # You'll have to define "profilesEditPassword" in secrets.yml, or this will fail. Thanks, ansible.
-  http_basic_authenticate_with name: Rails.application.secrets.profiles_edit_user, password: Rails.application.secrets.profiles_edit_password, except: [:index, :show]
+  UNITS = { access: 'Access Services', archives: 'Archives', augustana: 'Augustana Library',
```
pgwillia: YASS!  But there are still a couple of mentions of `$units` in https://github.com/murny/discovery/blob/rubocop-fixes-attempt/app/views/profiles/show.html.haml#L32, https://github.com/murny/discovery/blob/rubocop-fixes-attempt/app/views/profiles/index.html.haml#L289 and https://github.com/murny/discovery/blob/rubocop-fixes-attempt/app/views/profiles/index.html.haml#L331

app/controllers/profiles_controller.rb
```diff
@@ -1,88 +1,111 @@
...
+            bib: 'Bibliographic Services', saint: 'Bibliothèque Saint-Jean', business: 'Business Library',
+            collections: 'Collection Strategies', digital: 'Digital Initiatives',
+            education: 'Education / Kinesiology, Sport & Recreation Library', facilities: 'Facilities',
+            finance: 'Financial Systems & Analysis', health: 'Health Sciences Library', hr: 'Human Resources',
+            humanities: 'Humanities & Social Sciences / Law Libraries', its: 'Information Technology Services',
+            admin: 'Learning Services/Libraries Administration', science: 'Science & Technology Library',
+            special: 'Special Collections', press: 'University of Alberta Press' }.freeze
+
+  BUILDINGS = { augustana: 'Augustana Campus Library', bsj: 'Bibliothèque Saint-Jean',
```
pgwillia: Again, so much YASS! But https://github.com/murny/discovery/blob/rubocop-fixes-attempt/app/views/profiles/index.html.haml#L338 and https://github.com/murny/discovery/blob/rubocop-fixes-attempt/app/views/profiles/show.html.haml#L26

app/services/symphony_service.rb
```diff
@@ -42,119 +38,132 @@ def items
...
 
   def node_text(current_node)
-    current_node.at_xpath(".//xmlns:text", :xmlns=>"http://schemas.sirsidynix.com/symws/standard").text
+    current_node.at_xpath('.//xmlns:text', xmlns: 'http://schemas.sirsidynix.com/symws/standard').text
   end
 
-  def valid? id
-    (id =~ /^[0-9]*$/) == 0
+  def valid?(id)
+    /^[0-9]*$/.match?(id)
```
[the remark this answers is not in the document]
murny: Oooh! Good fine! Can revert back

Guess we will leave it the way it is. (tricky part with this is `id` here can be null, which leaves rubocops recommend way of using `.zero?` not possible)

app/controllers/profiles_controller.rb
murny: Good catches! Will refactor. 

View shouldn't be accessing these directly. Will create a instance variable like we do for new/edit and use these instead in the views.

```
    @buildings = BUILDINGS
    @units = UNITS
```

PR #1376, opened 2018-11-06
remove unused gems

pgwillia:
jbuilder and coffee-rails don't appear to be used.  Removing them!
We shouldn't be using therubyracer in production either... https://devcenter.heroku.com/articles/rails-asset-pipeline#therubyracer rails/rails#29285

closes #1375


pgwillia:
I'll send @nmacgreg an email/helpdesk ticket right now with a heads up.  I'll also open a deployment issue in github to track this change.

Issue #1259, opened 2018-09-10
[Discovery] Implement BPSC No Loan Form


weiweishi:
Once the wireframe is finalized here, #1216, the form needs to be developed on the front end:

https://balsamiq.cloud/s3awdua/pmnl7qx/rE1BA

**Requirements**
Auto-populated fields:

- the title of the object
- call number of the object.

URL for this form is:
 https://library.ualberta.ca/bpsc-readonsite

The original request was that these forms need to be sent to Jeff Papineau jeff.papineau@ualberta.ca and Kevin Zak kzak@ualberta.ca (@seanluyk I'd like to push back on this requirement, as using personal email is not a good idea not only because the communications only go to individual emails, but this also means that any personnel change would need a deployment of the application. )

seanluyk:
@weiweishi  - perhaps they have a delegated email account they could use instead? I'll check.  

For the URL on this and the RCRF read on site form, my preference would be to use the following convention:  /libraryacronym-location, e.g. /bpsc-readonsite

weiweishi:
@seanluyk They do currently have a delegated email but I think that copies to everyone in BPSC, and they don't want that. If they'd prefer not to use the existing one, I wonder if we could suggest them to create a new one and set cc to these personal emails that they'd need?

weiweishi:
Thanks @seanluyk I've updated the requirement for URL to be https://library.ualberta.ca/bpsc-readonsite

murny:
Wireframe/Mockup is now finalized and can be found here: https://github.com/ualbertalib/discovery/issues/1246

seanluyk:
@murny I've sent an email to BPSC staff. I'll let you know what they say. I like the idea of a newly created delegated account (they don't want to use their generic account).

seanluyk:
Adding another requirement to this ticket that's come out of testing:

When a user submits a request form, BPSC staff would like a confirmation email indicating that the request was received to be submitted. Specific text to follow soon

seanluyk:
In production, feature flag removed. Closing.

Issue #1258, opened 2018-09-10
Implement Micro/Newspaper Request form


weiweishi:
Once the wireframe is finalized here, #1216, the form needs to be developed on the front end:

https://balsamiq.cloud/s3awdua/pmnl7qx/r8F6C

**Requirements**
Auto-populated fields:

- Title
- URL to the record 

The request form should be sent to depository@ualberta.ca, and ill@ualberta.ca

The request url would be  https://library.ualberta.ca/rcrf-specialrequest

theLinkResolver:
@weiweishi @seanluyk I don't want to speak out of turn on this matter, but with respect to the name for the URL, it was once hoped that eventually the unbarcoded newspapers would be processed and the need to request them this way would disappear. (I presume we're talking about [these sorts of things](https://www.library.ualberta.ca/catalog/2687592))

theLinkResolver:
...so that is to say that the newspaper aspect may diminish in importance over time, which may affect the naming convention :)

seanluyk:
Good point, @theLinkResolver. In that case, for this form, how about we go with /rcrf-specialrequest or something as a catch-all?

weiweishi:
@theLinkResolver @seanluyk thanks. I've updated the requirements for this form, to use:
 https://library.ualberta.ca/rcrf-specialrequest

seanluyk:
As the form is in production under a feature flag, this work is done so the issue can be closed

PR #1386, opened 2018-11-08
RCRF/BPSC Request Forms

murny:
#1258
#1259 
#1260 

New forms will live at the following urls (once live):

http://library.ualbert.ca/bpsc_read_on_site_requests/new
http://library.ualbert.ca/rcrf_read_on_site_requests/new
http://library.ualbert.ca/rcrf_special_requests/new

And passing query params of `item_url` or `title` at the end of the url, will prepopulate the forms item url and title fields.

TODO:
- [x] routes
- [x] models
- [x] controllers
- [x] wrap up views
- [x] Add content for mailers
- [x] I18n text
- [x] tests
- [x] Get title from document passed to forms (needs refactoring of business logic in views...)

Future questions/work:
- Work is being done to get a general email for BPSC requests (currently we using two staffs personal emails)
- There is no link for the RCRF Special Requests form on the website?

Screenshots:
image
image
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'devise'
 gem 'devise-guests'
 
 # Misc Utilities
+gem 'bootstrap_form', '~> 2.7.0'
 gem 'friendly_id'
 gem 'nokogiri', '~> 1.8.5'
 gem 'om'
--- a/app/assets/stylesheets/ualib-discovery.scss
+++ b/app/assets/stylesheets/ualib-discovery.scss
@@ -420,6 +420,30 @@ h5 a, h5 a:visited, h5 a:hover,  h5 a:active{
   background-color: #e3f1f1;
 
 }
+
+// FIXME:
+// lightboxLink is a special class which has javascript associated with it
+// (pops up the link in a modal via blacklight's ajax modal code)
+// Hack this to get similar styling but without the javascript functionality
+// Ideally should just be using bootstrap btn instead of custom code
+// And css should be dashed cased and not camel cased
+#holdings table a.normal-link{
+  display: block;
+  background-color: #fff;
+  padding: 3px;
+  text-align: center;
+  border-radius: 5px;
+  margin-bottom: 3px;
+  text-decoration: none;
+  border:1px solid #bfc0c1;
+  font-size: 14px;
+  width: 70px;
+}
+#holdings table a.normal-link:hover{
+  background-color: #e3f1f1;
+
+}
+
 .holding-status, .holding-type{
   text-transform: capitalize;
 }
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -41,8 +41,7 @@ module HoldingsHelper
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
 
   def read_on_site_path(item, document_id, document_title)
-    # TODO: Add a feature flag for this feature for time being (remove in future)
-    return unless item[:status] == 'READONSITE' && params[:feature_flag_request_forms].present?
+    return unless item[:status] == 'READONSITE'
 
     if item[:location] == READ_ON_SITE_LOCATION_RCRF
       new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -42,12 +42,12 @@ module HoldingsHelper
 
   def read_on_site_path(item, document_id, document_title)
     # TODO: Add a feature flag for this feature for time being (remove in future)
-    if item[:status] == 'READONSITE' && params[:feature_flag_request_forms].present?
-      if item[:location] == READ_ON_SITE_LOCATION_RCRF
-        new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
-      elsif item[:location] == READ_ON_SITE_LOCATION_BPSC
-        new_bpsc_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
-      end
+    return unless item[:status] == 'READONSITE' && params[:feature_flag_request_forms].present?
+
+    if item[:location] == READ_ON_SITE_LOCATION_RCRF
+      new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
+    elsif item[:location] == READ_ON_SITE_LOCATION_BPSC
+      new_bpsc_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
     end
   end
 
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -40,13 +40,11 @@ module HoldingsHelper
   READ_ON_SITE_LOCATION_RCRF = 'UARCRF'.freeze # Research and Collections Resource Facility
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
 
-  # returns path to read on site form
-  # TODO: Should be passing in title as well, need to work out business logic for this though
-  def read_on_site_path(item, document_id)
+  def read_on_site_path(item, document_id, document_title)
     if item[:location] == READ_ON_SITE_LOCATION_RCRF
-      new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id))
+      new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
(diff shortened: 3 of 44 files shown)


Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'devise'
 gem 'devise-guests'
 
 # Misc Utilities
+gem 'bootstrap_form', '~> 2.7.0'
```
murny: We already have this in our gemfile via comfortable_mexican_sofa...so lets actually add it to ours as well and use it!

app/assets/stylesheets/ualib-discovery.scss
```diff
@@ -420,6 +420,30 @@ h5 a, h5 a:visited, h5 a:hover,  h5 a:active{
   background-color: #e3f1f1;
 
 }
+
+// FIXME:
```
murny: As comment says, the `lightboxLink` class we were using does some hidden weirdness. It will try to open the form into a modal, which breaks horrible as these forms are much much bigger then what the modal is meant for.

This is a great example of why you never style and have javascript for the same classes.

As a result, I want the same style, but none of the javascript....which means I need to copy and paste this over into a new class :slightly_frowning_face: 

Ultimately this button has the same css as the `btn btn-default` from bootstrap except for green color text. Probably could be some refactoring to be had here

(also renamed to `normal-link` instead of `normalLink` as css should be dashed cased)

app/models/solr_document.rb
```diff
@@ -28,4 +28,18 @@ class SolrDocument
   # and Blacklight::Solr::Document#to_semantic_values
   # Recommendation: Use field names from Dublin Core
   use_extension(Blacklight::Document::DublinCore)
+
+  # Example of how to move this business logic into the model...
```
murny: Exercise left for future developers I guess.

But this is an example of how we can take common things like getting the `title` out of the document object and handling all the edge cases and pushing this all down into the model, so you don't have to repeat this same logic all over the code base.

app/views/catalog/_authors.html.erb
```diff
@@ -1,9 +0,0 @@
-
```
murny: Never called now that I removed the commented out partial call on the show page. Delete it

app/views/catalog/_show_header_default.html.erb
```diff
@@ -1,11 +1,5 @@
...
-  <% @display_title = document.to_h["title_display"] %>
-<% elsif document.to_h["title_display"].class == Array %>
-  <% @display_title = document.to_h["title_display"].first %>
-<% end %>
 <h1>
-    <%= "#{@display_title}" %><%= ", #{document.to_h["edition_tesim"].first}" if document.to_h["edition_tesim"]%>
+    <%= document.header_title %><%= ", #{document.to_h["edition_tesim"].first}" if document.to_h["edition_tesim"]%>
 </h1>
 <div id="eaccess">
-</div><br />
```
murny: We using `<br>` and `nbsp` all over the views. CSS exists for a reason.

config/initializers/inflections.rb
```diff
@@ -10,7 +10,7 @@
 #   inflect.uncountable %w( fish sheep )
 # end
 
-# These inflection rules are supported but not enabled by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.acronym 'RESTful'
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.acronym 'RCRF'
+  inflect.acronym 'BPSC'
```
murny: Because these are acronyms, we want them to say this way. So when you use various rails methods, sometimes these may get displayed weirdly. 

For example, before this, the human version of `RCRFSpecialRequest` as a title/string was `Rcrf special request`, this keeps it to `RCRF special request`

app/mailers/request_form_mailer.rb
```diff
@@ -0,0 +1,19 @@
+class RequestFormMailer < ApplicationMailer
+  def bpsc_read_on_site_request(bpsc_read_on_site_request)
+    @bpsc_read_on_site_request = bpsc_read_on_site_request
+    mail(to: ['jeff.papineau@ualberta.ca', 'kzak@ualberta.ca'],
```
murny: Once we get a new group email for BPSC, we can change this line here:
e.g:
`mail(to: 'bpscrequest@ualberta.ca',`

instead of the current emails here now.

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -45,12 +45,13 @@
                 <td></td>
               <% end %>
               <td>
-<% if false %>
-<%# TODO: this feature has not been completed %>
                   <% if item[:status] == 'READONSITE' %>
```
pgwillia:
```suggestion
<% if Rails.env.development? || Rails.env.uat? %>
                  <% if item[:status] == 'READONSITE' %>
```

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -45,12 +45,13 @@
...
-                    <%= link_to_if(read_on_site_path(item).present?, "Request Me", read_on_site_path(item), {:id => 'requestLink', :class => 'lightboxLink'}) %>
+                    <%= link_to_if(read_on_site_path(item, @document.id, @document.header_title).present?,
+                                   'Request Me',
+                                   read_on_site_path(item, @document.id, @document.header_title),
+                                   class: 'normal-link',
+                                   target: :_blank) %>
                   <%- end -%>
```
pgwillia:
```suggestion
                  <%- end -%>
<%- end -%>
```

app/views/catalog/_symphony_holdings.html.erb
murny: Going to create a real FF here using the params from the url. So by default it's turned off and only available  when url is something like:
`library.ualberta.ca/catalog/12345678?feature_flag_request_forms=1`

So people can still test this on UAT if they would like

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -45,12 +45,11 @@
                 <td></td>
               <% end %>
               <td>
-                  <% if item[:status] == 'READONSITE' %>
-                    <%= link_to_if(read_on_site_path(item, @document.id, @document.header_title).present?,
-                                   'Request Me',
-                                   read_on_site_path(item, @document.id, @document.header_title),
-                                   class: 'normal-link',
-                                   target: :_blank) %>
+                  <% if read_on_site_path(item, @document.id, @document.header_title).present? %>
```
murny: Move the `if item[:status]` check (and the new feature flag) to the helper, as this keeps our views a bit cleaner

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -45,12 +45,11 @@
                 <td></td>
               <% end %>
               <td>
-                  <% if item[:status] == 'READONSITE' %>
-                    <%= link_to_if(read_on_site_path(item, @document.id, @document.header_title).present?,
```
murny: This was a bug! Didn't quite understand `link_to_if`. So this renders the `Request Me` text regardless if this is true or false, it only makes the `Request Me` an actual link if the condition is true!

So this isn't what we wanted after all. Move the condition into a simple If clause instead.

app/helpers/holdings_helper.rb
```diff
@@ -41,10 +41,13 @@ def display_ual_shield(document)
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
 
   def read_on_site_path(item, document_id, document_title)
-    if item[:location] == READ_ON_SITE_LOCATION_RCRF
-      new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
-    elsif item[:location] == READ_ON_SITE_LOCATION_BPSC
-      new_bpsc_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
+    # TODO: Add a feature flag for this feature for time being (remove in future)
```
murny: These links/buttons will now be featured flagged and will only show to the user if they append a query param to their url.

e.g: `http://library.ualberta.ca/catalog/12345678?feature_flag_request_forms=1`

app/views/catalog/_symphony_holdings.html.erb
murny: On second revision, kept the feature flag here in the view, as this keeps our tests for the helper cleaner (as a helper is just a ruby module and probably shouldn't have any context with regards to request/response lifecycle (params object)

murny:
Documenting the feature flag might be a bit overkill for this as it's hopefully temporary. But definitely should get added to the 1258-1260 issues when we tell the stakeholder to go test this

app/mailers/request_form_mailer.rb
seanluyk: @murny the request email has been created

Issue #877, opened 2016-01-14
[Discovery] some URLs don't display


theLinkResolver:
Some URLs (based on the 856 field indicators, I believe) do not display. Special Collections noted that this means that the links to finding aids are not displaying the record, for example:

https://www.library.ualberta.ca/catalog/6282297
http://www.library.ualberta.ca/permalink/opac/6282297/WUAARCHIVE

856 indicators are 42, with the 2 meaning "related resource" (i.e. a link that is not for _the_ resource itself).

This coding is also used for links to supplementary materials, publisher's descriptions, etc.

Done when:
- [ ] Finding aid links displaying in the records

redlibrarian:
@theLinkResolver Blacklight doesn't look at the indicators, so the presence of a label and url should be sufficient. The XML in a record that does display the link properly is the same as the XML in this record:

    <MarcEntryInfo>
    <label>Electronic access</label>
    <entryID>856</entryID>
    <indicators>40</indicators>
    <text>University of Alberta Access: 1985-</text>
    <unformattedText>
    |3University of Alberta Access: 1985- |uhttp://resolver.library.ualberta.ca/resolver?    ctx_enc=info%3Aofi%2Fenc%3AUTF-8&ctx_ver=Z39.88-    2004&rfr_id=info%3Asid%2Fualberta.ca%3Aopac&rft.genre=journal&rft.object_id=954925545272&rft.i    ssn=0820-3946&rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Ajournal&url_ctx_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Actx&url_ver=Z39.88-2004
    </unformattedText>
    <entryTypeCodes>iL</entryTypeCodes>
    <url>
    http://resolver.library.ualberta.ca/resolver?ctx_enc=info%3Aofi%2Fenc%3AUTF-8&ctx_ver=Z39.88-2004&rfr_id=info%3Asid%2Fualberta.ca%3Aopac&rft.genre=journal&rft.object_id=954925545272&rft.issn=0820-3946&rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Ajournal&url_ctx_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Actx&url_ver=Z39.88-2004
    </url>
    </MarcEntryInfo>

    <MarcEntryInfo>
    <label>Electronic access</label>
[...]

PR #1398, opened 2018-11-20
add supplimentary urls (such as finding aids) to electronic access div

pgwillia:
#877 
screenshot from 2018-11-20 15-54-56
screenshot from 2018-11-20 15-54-25
screenshot from 2018-11-20 15-53-49


app/assets/stylesheets/ualib-discovery.scss
```diff
@@ -278,6 +278,7 @@ font-size: 18px;
 }
 #electronic_access{
   font-size: 16px;
+  padding: 25px;
 }
 #electronic_access span{
```
pgwillia: electronic access link text was wrapped in `<span> ` to get this styling (smaller font).  I've not put in the effort to add the span.  Need to investigate what the difference between `electronic_access span` and `electronic_acess a`

app/helpers/catalog_helper.rb
```diff
@@ -0,0 +1,17 @@
+module CatalogHelper
+  include Blacklight::CatalogHelperBehavior
+
+  def electronic_access_label(location)
+    loc = location.split(":").last.strip.gsub("University of Alberta Access", "").strip
```
pgwillia: Copied from erb template.  Could refactor.

seanluyk:
Excellent to see progress on this important issue, @pgwillia!

app/views/catalog/_main_url_display.html.erb
```diff
@@ -1,22 +1,13 @@
...
-      <% end %>
-      <div class=\"row\"> \
-    <a href=\"<%= @proxy+url %>\" target=\"_blank\"><span>Click Here for University of Alberta Access <%= loc.blank? ? "" : "("+loc+")" %></span><span></a><% end %><% end %><br />";
-var urls_node = document.createElement('div');
-urls_node.setAttribute('id', 'electronic_access');
-urls_node.innerHTML = urls_html;
-node = document.getElementById("eaccess");
-node.appendChild(urls_node);
-</script>
+<div id='electronic_access'>
```
murny: This is great! Nice refactor! :100: 

My concern I guess is what about the `_database.html.erb` and `_urls.html.erb` which are doing very similar things here? Be nice if we can clean these up too. (since, if they do get rendered to the same page, this can have some weird side effects (ones doing javascript, other is erb rendered) and having multiple IDs that are the same is bad! As IDs should be unique)

Looks like all three are being called from the `show.html.erb` page here:

https://github.com/ualbertalib/discovery/blob/05ee81a0349bf9d315bb2cc6faed2c5c2c91291c/app/views/catalog/show.html.erb#L49-L60

We probably could take your refactor and apply it to these other ones and maybe just have one partial instead of three! (like half of `_urls.html.erb` will NEVER be rendered as `@ual_urls`  will always be empty since we check this before rendering the partial :thinking: )

app/views/catalog/show.html.erb
```diff
@@ -50,9 +50,6 @@
     <%= render "databases" %>
       <% end %>
 
-      <% if @ua_urls %>
-        <%= render "main_url_display" %>
-      <% end %>
       <% if @non_ua_urls %>
         <% if @ua_urls.empty? %>
           <%= render "urls" %>
```
murny: This is great! Nice refactor! :100: 

My concern I guess is what about the `_database.html.erb` and `_urls.html.erb` which are doing very similar things here? Be nice if we can clean these up too. (since, if they do get rendered to the same page, this can have some weird side effects (ones doing javascript, other is erb rendered) and having multiple IDs that are the same is bad! As IDs should be unique)

The only place these partials get called is right here in the `show.html.erb` page here (L46-L57)

We probably could take your refactor and apply it to these other ones and maybe just have one partial instead of three! (like half of `_urls.html.erb` will NEVER be rendered as `@ual_urls`  will always be empty since we check this before rendering the partial :thinking: )

pgwillia: This made my head hurt, but it's done.  To the best of my knowledge it does the same things as before.

app/views/catalog/_show_header_default.html.erb
```diff
@@ -4,12 +4,58 @@
...
+              <% loc=loc.delete('()').strip %>
+          <% end %>
+          <div class="row">
+            <a href=\"<%= @proxy+url %>\" target=\"_blank\"><span>Click Here for University of Alberta Access <%= loc.blank? ? "" : "("+loc+")" %></span><span></a>
+          </div>
+        <% end %>
+      <% end %>
     <% elsif @non_ua_urls %>
-      <%= render 'urls' %>
+      <% unless @ua_urls.empty? %>
```
pgwillia: I don't think this block is reachable?!

app/helpers/catalog_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+  def database_electronic_access?
+    @document['format'].include?('Database') && @document['url_tesim'].present?
+  end
+
+  # These are typical electronic access links and appear at the top of the page
+  # for example
+  # (Database) catalog/11709037
+  # (University of Alberta Access) catalog/2566934
+  # (Finding Aid) catalog/4072071
+  def electronic_access_top?
```
murny:
```suggestion
  def electronic_access_top?
    database_electronic_access? || @ua_urls.present? || supplimentary_urls.present?
  end
```

This would do the exact same thing as your code.

Another way to think of it is: 

```
 if database_electronic_access? || @ua_urls.present? || supplimentary_urls.present?
    true
  else
    false
  end
```

app/views/catalog/_databases.html.erb
```diff
@@ -1,9 +0,0 @@
-<script language="javascript">
-var urls_html = '<br /><table><tr></tr><% if @document["url_tesim"] %><tr><td><a href="<%= @document["enableproxy_tesim"].first == "1" ? Rails.application.config.proxy : "" %><%= @document["url_tesim"].first %>" target=\'_blank\'>Click Here for Access</a></td></tr><% end %></table><br />';
```
pgwillia: Left out these tables because I didn't think it added anything.

app/helpers/catalog_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+  def database_electronic_access_url(document)
+    electronic_access_url(document['enableproxy_tesim'].first == '1', document['url_tesim'].first)
+  end
+
+  def electronic_access_url(test, url)
+    return url if test
+    Rails.application.config.proxy + url
+  end
+
+  def supplimentary_urls
```
pgwillia: Not sure if this should be here or a variable assigned in the controller.

app/helpers/catalog_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+    database_electronic_access? || @ua_urls.present? || supplimentary_urls.present?
+  end
+
+  # These are 'On-campus access' and appear at the bottom of the page
+  # for example catalog/6990969
+  def electronic_access_bottom?
+    !electronic_access_top? && @non_ua_urls
+  end
+
+  def eaccess_label(location)
```
pgwillia: This was a partial before. Kept the name but perhaps it could be more descriptive.

app/views/catalog/show.html.erb
```diff
@@ -43,27 +43,23 @@
...
-        <br />
-        <%= "No copies found. This item may be lost or missing, or the data may be temporarily unavailable. Please <a href='/ask-us' target='_blank'>ask us</a> for assistance.".html_safe %>
-      <% end %>
-
-      <% else %>
-  <%= render "holdings" %>
-<% end %>
+    </div>
+  <% end %>
+  <% if @holdings.nil? || @holdings.empty? %>
```
pgwillia: Changed spacing at the beginning of line so that this is easier to read.

app/views/catalog/_show_header_default.html.erb
```diff
@@ -1,5 +1,20 @@
 <h1>
     <%= document.header_title %><%= ", #{document.to_h["edition_tesim"].first}" if document.to_h["edition_tesim"]%>
 </h1>
-<div id="eaccess">
```
pgwillia: I think eaccess was just a placeholder for the JavaScript to find. Doesn't add anything now so removed.

app/views/catalog/_databases.html.erb
```diff
@@ -1,9 +0,0 @@
...
-var urls_node = document.createElement('div');
-urls_node.setAttribute('id', 'electronic_access');
-urls_node.innerHTML = urls_html;
-var page_id = "doc_"+<%= @document["id"] %>;
-node = document.getElementById(page_id).getElementsByTagName("h1")[0];
```
pgwillia: Not exactly in the same place as before but now same as others.

app/views/catalog/show.html.erb
murny:
```suggestion
  <% if @holdings.blank? %>
```

Can use `.blank?` here as this does the same thing (this helper comes from rails ActiveSupport and isn't found in Ruby standard library).

So given this example you can see the difference:

```
2.5.3 :001 > holdings = nil
 => nil 
2.5.3 :002 > holdings.empty?
Traceback (most recent call last):
        1: from (irb):2
NoMethodError (undefined method `empty?' for nil:NilClass)
2.5.3 :003 > holdings.blank?
 => true 
2.5.3 :004 > holdings.nil?
 => true 
2.5.3 :005 > holdings = []
 => [] 
2.5.3 :006 > holdings.empty?
 => true 
2.5.3 :007 > holdings.nil?
 => false 
2.5.3 :008 > holdings.blank?
 => true 
```

pgwillia: I was considering that outside of the scope of what I was looking at.  This is logic to display the holdings table rather than any electronic access link.  I've made the change anyways 'cause it's a good suggestion.

PR #1408, opened 2018-11-30
#1340: Band aid fix and clean up some of the partials for email record

murny:
#1340 

So from looking at the stacktraces, I believe this could be a loading issue/race condition issue with how the blacklight gem files come in and our overrides come in mainly around the `include Blacklight::Category`. Which we are overriding this entire file in discovery to handle some of our own usecases (found in `lib/blacklight/category.rb`). For whatever reason, for these errors, it was running the blacklight code instead of our own overrides.

So in this case, we have defined some extra things in the email form that is not standard in blacklight. Such as the ability to have the Call Number and Location from the holdings table be shown in the email body as well as you can email these records to yourself. But forwhatever reason, it is running the old code from blacklight. So when it gets to the email body template, it tries to print out the Call/Location which should exist or at the very least be an array. However since it's running the blacklight code, these things are actually nil. So you can't call `nil.empty?` as this raises an exception. So instead, if there is ever a case where it is running the old code, we can safe guard ourselves by calling `nil.blank?` instead. Does the same thing, just protects yourself if this input is `nil` or an `[]`.

There is a ton of other issues and problems with this code, but I bit only enough to fix this root issue and update some of the templates. Ideally we should be doing a proper monkey patching approach to get our code in and overwritting blacklight, instead of just hoping....our code gets picked up last. Plus tons of improvements here. (There is tons of logic to handle "Bookmarks" and emailing bookmarks to yourself which just complexes this code, you can go directly to this email form, and break it as it assume ID is always present, no validations except for email on the form, etc etc)


app/views/catalog/_email.html.erb
```diff
@@ -1,4 +0,0 @@
-
-        <li class="email">
-          <%= link_to t('Email me this item'), email_catalog_path(:id => @document), {:id => 'emailLink', :data => {:ajax_modal => "trigger"}} %>
```
murny: Just put this link where its being called... Don't need a 1 line partial for this

app/views/catalog/_email_form.html.erb
```diff
@@ -1,46 +0,0 @@
-<%= form_tag url_for(:controller => "catalog", :action => "email"), :id => 'email_form', :class => "form-horizontal ajax_form", :method => :post do %>
```
murny: Never used more then once and again, its hard to navigate this partial hell (since we have 70+ partials/views in /catalog). Let's just put this where it actually gets rendered out

app/views/catalog/email_sent.html.erb
```diff
@@ -1,9 +0,0 @@
-<div class="modal-header">
```
murny: This one is bad.... Firstly, this is never called as blacklight 5.15.0 has moved away from this partial. Here is how this partial looks like in blacklight 5.15.0:

```erb
<% Deprecation.warn controller, "The template catalog/email_sent.html.erb is deprecated; use email_success instead" %>
<%= render template: "catalog/email_success" %>
```

Okay, so you should be using `email_success` instead, which has this exact same contents as this file.

What calls `email_success.html.erb`? Nothing.

They use some metaprogramming...crazy...way of creating all these actions for the catalog (like bookmark/email/sms/librarian_view/etc etc etc)

As a result the important piece is the actual action and how they handle the "success" part. It looks like something like this:

```ruby
            flash[:success] ||= I18n.t("blacklight.#{name}.success", default: nil)

            respond_to do |format|
              format.html { redirect_to action_success_redirect_path }
              format.js { render "#{name}_success" }
            end
```
So if its a html response, we just redirect back to the catalog items show page and throw up a flash message. Thats great!

If its a JS response, we render in this case a `email_success` view! Okay, so maybe it will render the `email_success.html.erb` view?

Nope, what it will try to render is a `email_success.js.erb` view, since its a JS response it will look for a view with a JS extension not a HTML extension.

So this code never worked from the beginning... Thats fantastic.

So to sum it up, this file never gets rendered so why even have it?

app/views/record_mailer/email_record.text.erb
```diff
@@ -7,12 +7,12 @@ Here <%= @count == 1 ? "is the item" : "are the items" %> you emailed from the U
 <%= t('blacklight.email.text.url', :url =>polymorphic_url(document, @url_gen_params)) %>
 <%= "\n" %>
 <% end %>
-<% unless @call.empty? %>
+<% unless @call.blank? %>
```
murny: Here's the real quick fix, even if  discovery is running the blacklight code or discovery override, this will still work as ` .blank?`handlea both nil and empty array

Issue #1404, opened 2018-11-29
Add call number to the BPSC form


weiweishi:
This was not included in the initial wireframe but was raised when they reviewed the form on UAT. They'd like to have the Call Number of the item to be auto-populated into the form. 

Call number can be added between title and item url field on the form.

murny:
Call number is the thing thats highlighted here? Correct?

![image](https://user-images.githubusercontent.com/1930474/49254262-09f6d880-f3e6-11e8-9930-233aca69496a.png)

weiweishi:
Yes. That is call number.  And CJ didn't request to have this added to RCRF forms. I can double check with him to make sure. For now only need to add to BPSC form.

murny:
I made the assumption here in #1417 that this call number field on the form is a required field just like title and item url? Is this a fair assumption?

weiweishi:
I think this is an assumption that needs confirmation. I know BPSC wants to potentially use this form for requests that might not be triggered by Discovery records, and we will need to verify with them that all the requests they deal with are materials with call numbers.

murny:
@weiweishi Before we close this...Did we get confirmation about my assumption with regards to call number being a required field?

PR #1417, opened 2018-12-04
#1404: Add call number to BPSC On Site Form

murny:
#1404: Adds call number to request form for BPSC On Site.

Made the assumption that this is a required field?

screen shot 2018-12-04 at 9 42 15 am

### DIFF
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -44,14 +44,12 @@ module HoldingsHelper
   READ_ON_SITE_LOCATION_RCRF = 'UARCRF'.freeze # Research and Collections Resource Facility
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
 
-  def read_on_site_path(item, document_id, document_title)
-    return unless item[:status] == 'READONSITE'
+  def request_rcrf_read_on_site_form?(item)
+    item[:location] == READ_ON_SITE_LOCATION_RCRF && item[:status] == 'READONSITE'
+  end
 
-    if item[:location] == READ_ON_SITE_LOCATION_RCRF
-      new_rcrf_read_on_site_request_path(item_url: catalog_url(document_id), title: document_title)
-    elsif item[:location] == READ_ON_SITE_LOCATION_BPSC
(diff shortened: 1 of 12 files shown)


app/helpers/holdings_helper.rb
```diff
@@ -44,14 +44,12 @@ def display_ual_shield(document)
   READ_ON_SITE_LOCATION_RCRF = 'UARCRF'.freeze # Research and Collections Resource Facility
   READ_ON_SITE_LOCATION_BPSC = 'UASPCOLL'.freeze # Bruce Peel Special Collections
 
-  def read_on_site_path(item, document_id, document_title)
```
murny: Refactored this to do something very similar to `request_microform?`. Much simpler this way and keeps the link in the view.

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -47,18 +47,29 @@
...
                   <% if params[:feature_flag_request_forms].present? %>
-                    <% if request_microform? item %>
+                    <% if request_microform?(item) %>
                       <%= link_to('Request Me',
-                                  new_rcrf_special_request_path(item_url: catalog_url(@document.id), title: @document.header_title),
+                                  new_rcrf_special_request_path(item_url: catalog_url(@document.id),
+                                                                title: @document.header_title),
                                   class: 'normal-link',
                                   target: :_blank) %>
-                    <% elsif read_on_site_path(item, @document.id, @document.header_title).present? %>
```
pgwillia: Are these independent? Could we end up with multiple "Request Me" buttons?

murny: The only scenario where you could have more then one "Request Me" buttons is there's a chance to have both `RCRF Special Request` and `RCRF Read on Site Request` on the same item. 

This would only be possible if the location is the same (RCRF) and i suppose you could have `type` to be 'Microform' and `status` to be 'READONSITE'. Which not sure if this is even possible though but if it was you would get two buttons!

pgwillia: I think this is probably fine.  I'll approve and we can adjust if we get any further input.

PR #1418, opened 2018-12-04
#1334: Fix SMS model error

murny:
Very similar to #1340. Seems like something weird happen on the deployment. Our custom `lib/blacklight/catalog/` code wasn't being loaded and used, and instead the code from blacklight was being used instead.

Also from the stacktraces, you can see the app was running Blacklight v5.19.2 which is impossible if they were using our Gemfile? As we lock this to v5.15.0. So something definitely went weird when deployment happened. And something we should definitely keep an eye on in the future...How did we deploy with v5.19.2? Could also improve our code loading in via a proper monkey patch approach (initializer etc),  as mentioned in PR #1408.

image

So like PR #1408 we can do a quick fix by using `.blank` to handle this situation (if its nil or empty string) and prevent this from failing even if the wrong code gets loaded...

Also made refactors to cleanup the partials and other things while I was digging in here.

Random aside, the big reason we deviated from blacklight for the SMS/Texting feature was so we could add the Call Number for the holdings record as the Subject heading.

SMS when no call number is provided:
image

SMS when call number is provided:
image

Which looks bad? As we dropped the entire subject context ("From the UofA...", it's title, etc) and replace it with the call number (which is a random string of chars/nums). (We also deviated by removing alot of the additional information that blacklight was providing about the record in this email message)


app/views/catalog/_show_tools.html.erb
```diff
@@ -31,16 +31,13 @@
         <%= link_to 'Librarian View', librarian_view_catalog_path(id: @document), {:id => 'citeLink', :class => 'lightboxLink'} %>
       </li>
 
-      <% if @document.respond_to?(:to_email_text) %>
-        <li class="email">
-          <%= link_to t('Email me this item'), email_catalog_path(id: @document), {id: 'emailLink', data: {ajax_modal: "trigger"}} %>
-        </li>
-      <% end %>
-
-      <%- if @document.respond_to?( :to_sms_text ) -%>
```
murny: These `@document.respond_to?` are silly. They will always be true, as we define these right on the Solr Document model....Why even check this?

app/views/catalog/_show_tools.html.erb
```diff
@@ -31,16 +31,13 @@
...
-        <%= render "sms" %>
-      <%- end -%>
+      <li class="email">
+        <%= link_to t('Email me this item'), email_catalog_path(id: @document), {id: 'emailLink', data: {ajax_modal: "trigger"}} %>
+      </li>
 
+      <li class="sms">
+        <%= link_to t('Text me a link to this item'), sms_catalog_path(:id => @document), {:id => 'smsLink', :data => {:ajax_modal => "trigger"}} %>
```
murny: Similar to the email link partial.  Just move this here...silly to have a 1 line partial file lying around so you can render this here

app/views/catalog/sms.html.erb
```diff
@@ -2,4 +2,37 @@
   <button type="button" class="ajax-modal-close close" data-dismiss="modal" aria-hidden="true">×</button>
   <h1 class="modal-title"><%= t('blacklight.sms.form.title') %></h1>
 </div>
-<%= render :partial => 'sms_form' %>
+
+<%= form_tag url_for(:controller => "catalog", :action => "sms"), :id => 'sms_form', :class => "form-horizontal ajax_form", :method => :post do %>
```
murny: Just move the form partial here.

app/views/catalog/_symphony_holdings.html.erb
```diff
@@ -59,12 +59,8 @@
                                   target: :_blank) %>
                     <%- end -%>
                   <%- end -%>
-                  <% if @document.respond_to?( :to_email_text ) %>
-                    <%= link_to "Email Me", email_catalog_path(:id => @document, :call=>item[:call], :location=>library_location(item[:location])), {:id => 'emailLink', :class => 'lightboxLink'} %>
-                  <%- end -%>
-                  <%- if @document.respond_to?( :to_sms_text ) -%>
-                    <%= link_to "Text Me", sms_catalog_path(:id => @document, :call=>item[:call]), {:id => 'smsLink', :class => 'lightboxLink'} %>
-                  <%- end -%>
+                  <%= link_to "Email Me", email_catalog_path(:id => @document, :call=>item[:call], :location=>library_location(item[:location])), {:id => 'emailLink', :class => 'lightboxLink'} %>
```
murny: Again, remove these silly `@document.respond_to?` calls

app/views/catalog/sms_sent.html.erb
```diff
@@ -1,9 +0,0 @@
-<div class="modal-header">
```
murny: Same explanation here holds true for the `sms_sent` view: https://github.com/ualbertalib/discovery/pull/1408#discussion_r237985975

app/views/record_mailer/sms_record.text.erb
```diff
@@ -1,4 +1,3 @@
 <% @documents.each do |document| %>
-<%# document.to_sms_text %>
```
murny: Remove comment out code

lib/blacklight/solr/document/sms.rb
```diff
@@ -1,11 +0,0 @@
-# This module provides the body of an email export based on the document's semantic values
-module Blacklight::Solr::Document::Sms
-  # Return a text string that will be the body of the email
-  def to_sms_text
-    # locations ||= YAML.load_file("#{Rails.root}/config/locations.yml")
```
murny: Since our app never uses this (only place its called in blacklight/discovery is in the `sms_record.text.erb` above), we don't need to override this just to comment out, we can just leave this in blacklight how it is. Remove this from Discovery.

app/models/record_mailer.rb
```diff
@@ -18,7 +18,7 @@ def email_record(documents, details, url_gen_params)
   def sms_record(documents, details, url_gen_params)
     @documents      = documents
     @url_gen_params = url_gen_params
-    subject = details[:call].empty? ? "From University of Alberta Libraries: #{documents.first.to_semantic_values[:title].first}" : details[:call]
```
murny: Problem was when running the code from blacklight, `details[:call]` is nil since it's not being passed to our `sms_record` from the `sms_action`. Hence `nil.empty?` will error out. If we change this to `.blank?` then this will work even if `details[:call]` is an empty string or a nil object.

PR #1424, opened 2018-12-11
sanitize notes in contents field

### LINKED ISSUE #1085: [Discovery] quotes in contents field being rendered as &quot;
This must have something to do with the formatting that is applied to the contents field, as quotes look okay elsewhere in the record display.

Examples:
https://www.library.ualberta.ca/catalog/7583704
https://www.library.ualberta.ca/catalog/7583375

pgwillia:
screenshot from 2018-12-11 14-25-48

fixes #1085

### DIFF
--- a/app/views/catalog/_show_default.html.erb
+++ b/app/views/catalog/_show_default.html.erb
@@ -27,8 +27,7 @@
         <dd class="blacklight-<%= solr_fname.parameterize %>" style="font-size:75%">
         <ul>
           <% render_document_show_field_value(document,:field=>solr_fname).split("--").each do |note| %>
-
-            <li><%= note.gsub("&#39;", "'") %></li>
+            <li><%= sanitize note %></li>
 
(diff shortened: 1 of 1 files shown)


app/views/catalog/_show_default.html.erb
```diff
@@ -27,8 +27,7 @@
         <dd class="blacklight-<%= solr_fname.parameterize %>" style="font-size:75%">
         <ul>
           <% render_document_show_field_value(document,:field=>solr_fname).split("--").each do |note| %>
-
-            <li><%= note.gsub("&#39;", "'") %></li>
+            <li><%= sanitize note %></li>
```
murny: Should always use parenthesis when invocating methods with args like in this case:
```suggestion
            <li><%= sanitize(note) %></li>
```

More info: 
https://github.com/rubocop-hq/ruby-style-guide#method-invocation-parens

PR #1429, opened 2018-12-13
solve permissions issue on log and tmp directory

pgwillia:
`rake assets:precompile` in the Dockerfile was changing the owner of logs and tmp.  Same as https://github.com/phusion/passenger-docker/issues/143#issuecomment-234171238

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,10 @@ COPY Gemfile.lock $APP_ROOT
 RUN bundle install --without development test --jobs=3 --retry=3
 
 # *NOW* we copy the codebase in
-COPY --chown=app:app . $APP_ROOT
-
+COPY . $APP_ROOT
 # Precompile Rails assets.
 RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
+# change the owner so that the app passenger_user can read and write to the app dir
+RUN chown -R app:app .
 
 EXPOSE 80


Dockerfile
```diff
@@ -37,5 +37,6 @@ COPY --chown=app:app . $APP_ROOT
 
 # Precompile Rails assets.
 RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
+RUN chown -R app:app .
```
murny: Do we need the chown -R app:app on the copy files over now? https://github.com/ualbertalib/discovery/blob/617b982ea5adf033cbcf93cef858cf6719b5ea17/Dockerfile#L36

Since we also do it here because of assets:precompile? Or does it matter that we do it twice 🤔?

pgwillia: Good point.  It could be:
```
# don't set the owner yet
COPY . $APP_ROOT
RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
# set it here
RUN chown -R app:app .
```
or
```
COPY --chown=app:app . $APP_ROOT
RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
# just change the owner of the newly created log and tmp directories
RUN chown -R app:app tmp log
```
or maybe set the app user for the assets:precompile step?
```
COPY --chown=app:app . $APP_ROOT
# set the user
USER app
RUN RAILS_ENV=uat SECRET_KEY_BASE=pickasecuretoken bundle exec rake assets:precompile
# then you don't need to change the permissions
```

Which makes the most sense to you?  I haven't checked that all these work, but I think they would have the same outcome.

murny: Probably the first one is best? You want to chown all the files, but we can just do all at once at the very end. Instead of doing it for each step

PR #1431, opened 2018-12-14
Fix yanked rollbar version and update dependencies

murny:
Rollbar yanked 2.18.1 version from RubyGems, lets use the next version (problem occurred by merging  #1430 )

Also updated other dependencies since I was here....which shows we haven't been doing a good job with locking down our versions. So did this as well. Since we are stuck using Ruby 2.1.5...https://github.com/ualbertalib/discovery/issues/1432


Gemfile.lock
```diff
@@ -78,20 +78,21 @@ GEM
       blacklight (~> 5.15)
       jquery-rails
       rails (>= 3.0, < 5.0)
-    bootstrap-sass (3.3.7)
+    bootstrap-sass (3.4.0)
```
murny: Looks like we need to lock to bootstrap 3.3.7 since it requires ruby 2.3.3+....

Issue #1443, opened 2018-12-21
BPSC and RCRF form submission completion redirection behaviour


seanluyk:
**Describe the bug**
When either of the forms (BPSC or RCRF) are submitted, the users are redirected to the library homepage instead of the item that they were looking at.

**To Reproduce**
Steps to reproduce the behaviour:
1. Submit a form
2. Redirected to the home page

**Expected behaviour**
When a user submits a form, the completion message appears on the page they are viewing.

**Screenshot**
Here is a screenshot of the completion message appearing out of context (i.e. a search made after the fact):
![screen shot 2018-12-21 at 11 13 55](https://user-images.githubusercontent.com/14880206/50356999-9b52f980-0511-11e9-8f67-4e5bc47a6805.png)

murny:
This is a good find. The reason for this is the "homepage" comes from the CMS and for whatever reason doesn't support flash messages (green or red alerts like you see in your screenshot). So you only receive the flash message when you go back into a "blacklight" page as only these pages support them. (this isn't caught in development as we don't have a way to run the site with the CMS)

So easy fix, we can just add flash messages support to the CMS layout:
https://github.com/ualbertalib/discovery/blob/master/app/views/layouts/cms.html.erb#L55-L57

which they should have had to begin with just like the blacklight layout has them:
https://github.com/ualbertalib/discovery/blob/master/app/views/layouts/blacklight.html.erb#L48

As an aside,  why do we have 6 layouts in Discovery (should only need one)....
https://github.com/ualbertalib/discovery/tree/master/app/views/layouts


Regarding your comment on "the completion message does not appear in place. Instead, users are redirected to the library homepage.":

I would say this is pretty normal behaviour for websites? After the user fills out the form, they hit submit and are taken away from the previous form to the thing they just created (e.g: I deposit a thesis in ERA, after submitting the form, I am now on the thesis detail page). Even the Email/Text/Corrections forms in Discovery all work like this. After submitting my correction, I am redirected back to the catalog item page. So we should probably remain consistent with the rest of the website?

[...]

PR #1446, opened 2019-01-02
Add flash messages to CMS pages

murny:
fixes this issue: https://github.com/ualbertalib/discovery/issues/1443

- Removed 'blacklight' layout in favour of 'application' layout (nothing was using this...so lets get rid of one layout in favour of rails defaults)

Improvements:
- Why are we using 4 layouts for the CMS? Should be able to get away with maybe one? maybe two (cms and french cms?)? Read docs here on handling locale: https://github.com/comfy/comfortable-mexican-sofa/wiki/Docs:-Translations
- lots of duplication in these layouts (home/french home are not using the shared header partial etc)
- we got a 'wrapper' div...which closing tag is in the footer partial? And we have a `push` div...which isnt doing much at all...
- french layout has two google analytics scripts?

Above improvements captured in this ticket: https://github.com/ualbertalib/discovery/issues/1447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 ## [Unreleased]
 
 ### Fixed
-- Fix request forms success flash behaviour by adding flash to CMS layouts [#1446](https://github.com/ualbertalib/discovery/issues/1446)
+- Fix request forms success flash behaviour by adding flash to CMS layouts [#1443](https://github.com/ualbertalib/discovery/issues/1443)
 
 ## [3.0.94] - 2018-12-19
 


app/views/layouts/application.html.erb
```diff
@@ -1,3 +1,57 @@
...
+    <!--[if IEMobile]>
+      <meta http-equiv="cleartype" content="on">
+    <![endif]-->
+
+    <title>University of Alberta Libraries</title>
+    <%= opensearch_description_tag application_name, opensearch_catalog_url(:format => 'xml') %>
+    <%# favicon_link_tag asset_path('favicon.ico') %>
+    <%= favicon_link_tag 'favicon.ico' %>
+    <%= stylesheet_link_tag    "application", "https://fonts.googleapis.com/css?family=Open+Sans:400,700,400italic|Abel", "sticky-footer", "ualib-discovery", media: "all" %>
+    <%= javascript_include_tag "application", "stupidtable", "ualib-discovery" %>
```
pgwillia: was `<%= javascript_include_tag "application", "ualib-home" %>` Why this change?

app/views/layouts/application.html.erb
```diff
@@ -1,3 +1,57 @@
...
+
+  </head>
+  <body class="<%= render_body_class %>" data-no-turbolink="true">
+    <div class="wrapper">
+  <%= render :partial => 'shared/header_navbar' %>
+
+  <%= render partial: 'shared/ajax_modal' %>
+
+  <div id="main-container" class="container">
+    <%= render partial: '/flash_msg', layout: 'shared/flash_messages' %>
```
pgwillia: Which file does 'shared/flash_messages' refer to?

pgwillia: Ah!  It's in blacklight https://github.com/projectblacklight/blacklight/blob/v5.15.0/app/views/shared/_flash_messages.html.erb

murny:
Yeah good catches. Will add a changelog. 

Test be nice...but theirs some setup there (like getting someway to provision the CMS in the test environment etc? Heck this be nice to setup for development too!)

This will be tough and you bring up a good point. When library-cms gets pulled away, we will have to circle back and figure out a better way to handle this. Because they will then be two different apps they will not be able to share the flash messages in this way. So maybe a better approach would be instead of redirecting them to the `cms homepage` we can redirect them to the `catalog index`? Not sure... Needs some thinking for this transition (which is a bigger discussion about how the CMS and Discovery going to coexist as other things will need to change as well, etc)

app/views/layouts/application.html.erb
murny: This diff was me deleting `blacklight.html.erb` layout and moving everything over into `application.html.erb` as application.html.erb was never used and is rails default.

So this is just whatever blacklight.html.erb had.

`<%= javascript_include_tag "application", "ualib-home" %>` will include the javascript for both the application and ualib-home manifests. Which ideally should only be one javascript file... `application` not two. 

Some more refactoring could take place here.

app/views/layouts/application.html.erb
murny: Yep! I was in the progress of just using our own flash message partial... As exactly...where does this come from? Why we need a partial with a layout etc? but Blacklight references this flash partial for some of their javascript code. So figured better not change it in case it breaks a few things.

app/views/layouts/application.html.erb
pgwillia: I must have done something funny when I was looking at a diff of this file and blacklight.html.erb. No concerns here

murny:
Changelog entry added :+1:

Issue #1330, opened 2018-10-17
[Discovery] More complete feature test suite


pgwillia:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I try to make changes, or do updates and I'm not confident that I've checked every possible regression, or even know what all the features are to test.

**Describe the solution you'd like**

Results
- [x] Shield/logo to represent UofA holdings
- [x] Google book jackets
- [x] Metadata for symphony holdings includes Author, format, publication year, copies owned by
- [x] Metadata for ejournals and databases is more limited.
- [x] By default 25 results are displayed.
- [x] Many more facets

Record View
- [x] Go to bookmarks (#1562)
- [x] Send a correction (#1727)
- [x] View search history (#1725)
- [x] Text me (Canadian SMS) / email me (#1722)
- [x] Journal records contain proper metadata (#1704)
- [x] Chained subject linking (#1708)
- [x] Author link on record view page search for the author (#1710)
- [x] Content presented in the holding table is proper (#1712)

Databases and ejournals Record View
- [x] Usage rights in the holding table (#1712)
- [ ] Coverage information (blocked by #1473)
- [x] A tiny link to ILL (#1740)

Search Result
- [X] Create feature tests for the search result page: https://github.com/ualbertalib/discovery/issues/1600

Logging (tests will be implemented in #423)
- [ ] errors in logs
- [ ] deprecations in logs

Fix blacklight/spec/rails helper https://github.com/ualbertalib/discovery/pull/1464#discussion_r248857540

Many of these customizations are identified in this report: https://docs.google.com/document/d/1VTeu4ZYfLhbLWsznRqFBU0qccQhD17J83tW6LDQu7OY/

**Describe alternatives you've considered**
Continue to test manually and have low confidence when deploying.

seanluyk:
@pgwillia this is really fantastic, I fully support this work given our current deployment issues

murny:
NOTE: as part of this ticket we need to get off ingesting tons of objects into solr on every time you run the tests? You don't need thousands of objects to test the code. 

As a result this makes the tests take far longer then they should be (rspec itself takes no time at all to run (since we only have 75 tests), but this ingestion process takes 3 times longer then the actual rspec process). And also this makes debugging travis difficult because of the spam the ingestion creates and as a result you can't see anything in travis about the actual tests: 
![image](https://user-images.githubusercontent.com/1930474/47525740-39e91400-d85b-11e8-901e-dd6d288356b1.png)

[...]

PR #1464, opened 2019-01-17
Use vcr for external services

### LINKED ISSUE #1463: Use VCR for tests that hit external services
**Is your feature request related to a problem? Please describe.**
Today [one of our tests](https://travis-ci.org/ualbertalib/discovery/builds/480989386?utm_source=github_status&utm_medium=notification) was failing because the test record became stale (no longer part of our subscription). 

**Describe the solution you'd like**
If we use vcr to record these transactions then we'll no longer have this issue.

**Describe alternatives you've considered**
>should honestly just delete all the tests in discovery and call bankruptcy :sweat_smile: (with its crazy ingest stuff, tests that really provide no value (a == a), and tests that depend on external services)

**Additional context**
We use vcr in our other rails projects https://github.com/ualbertalib/jupiter/blob/add9c58c6b8c9642931a2e58e296c8ce3bd923c6/test/test_helper.rb#L13-L19
`VCR.use_cassette('ezid_minting', erb: { id: item.id }, record: :none) do`

pgwillia:
closes #1463 

Testing is failing because the record we use to test the electronic holdings no longer has any holdings. Falsified the marc xml to use a more current id of this item.  Added vcr to record use of external services.


pgwillia:
@seanluyk do you know if there is anything in the ws terms of service that prevents us from recording/storing `fixtures/vcr_cassettes/`?

spec/blacklight_helper.rb
```diff
@@ -6,6 +6,8 @@
 require 'active_support/inflector'
 require 'capybara/rspec'
 require 'capybara/rails'
+require 'vcr'
```
murny: No idea why we have this file?

The point of spec_helper vs rails_helper is spec_helper is suppose to be a lightweight version that doesn't need or depend on all the rails stuff. But here we bloat everything inside this blacklight_helper, then include it in spec_helper (where spec_helper shouldn't need all this stuff capybara/rails/etc)...which is then included in rails_helper.

No change required here. More so just documented this and linking this comment to this our testing refactor issue https://github.com/ualbertalib/discovery/issues/1330

Issue #1280, opened 2018-09-22
Refactoring of Catalog view partials


weiweishi:
Catalog views folder needs a heavy refactor to reduce the number of views and partials that are short files (1-3 lines) that are never used more than once. We could also deal with the i18n issue as part of the refactoring work.

murny:
Some quick thoughts about working on this:

Can probably start with reviewing the view files (show.html.erb, index.html.erb, etc.) and anytime a render partial is called you can stop and evaluate it's usefulness. 

Does this really need to be a partial? Why is this a partial? Is this partial called more then once, aka its being used to keep the code DRY? If not, its probably okay to just remove this partial and put its code to where its being rendered. Majority of these partials are 1-3 lines long and we can safely just remove them after putting the code where it needs to be.

For the partials left, we can just go through these separately. Some of these are probably never used like this footer partial which is horribly out of date and all links point no where: https://github.com/ualbertalib/discovery/blob/master/app/views/catalog/_footer.html.erb

Then we can fix the html/erb itself! 

- Things like using I18n properly as noted above. 

- Most of these ERB files have horrible spacing/tabbing/alignment issues

- We also not using the rails helpers in a few places. Should be using these when possible instead of writing html.

- Removing lots of ruby logic from the views like variable assignment, complicated conditions, modifying the solr doc to get things like the docs title etc? (these should live in a helper or push them down into the model instead)

It might be worth bringing in ERB Lint (https://github.com/Shopify/erb-lint), which gives a nice way to lint ERB files like you would with rubocop (which will catch i18n issues, spacing/styling, etc and gives the ability to autofix theses (and keep them fixed as we could use this in travisCI and fail builds when violations happen))

After this you are probably looking pretty good! Be down to a few views and a small number of partials and all the code is starting to look pretty good!

This #1426 show page refactor be a great issue to pair with this view refactor

murny:
Think this is in pretty good shape now once #1492 is merged. Reduced partials from 57ish to 15. Cleaned up alot of the HTML/ERB code too.

Could be some more improvements but we can tackle this later when we do:
https://github.com/ualbertalib/discovery/issues/1476
https://github.com/ualbertalib/discovery/issues/1426
https://github.com/ualbertalib/discovery/issues/1448

PR #1465, opened 2019-01-17
Removing cms and profiles

pgwillia:
CMS moved to https://github.com/ualbertalib/library-cms  so profiles and cms can be removed from the discovery application.

#1280


config/routes.rb
```diff
@@ -40,10 +38,6 @@
   resources :bpsc_read_on_site_requests, only: [:new, :create]
   resources :rcrf_read_on_site_requests, only: [:new, :create]
 
-  root to: 'comfy/cms/content#show'
-
-  comfy_route :cms_admin, path: '/admin'
+  root to: 'catalog#index'
```
murny: Yay! We have a homepage again 🎊!

pgwillia:
@nnunn I'm removing the cms and profile pages from discovery and reclaiming the root page. I see that https://search-test.library.ualberta.ca/ goes to `/symphony` Is that the best default?  I was thinking `/catalog` should be. What do you think?

I thought that symphony was a specialized version of catalog like journals, ebooks and databases but it's not.  Is there any value to keeping it? We can [continue to use](https://guides.rubyonrails.org/routing.html#specifying-a-controller-to-use) 'symphony' in the url if necessary.

Gemfile
```diff
@@ -29,13 +29,8 @@ gem 'devise-guests'
 
 # Misc Utilities
 gem 'bootstrap_form', '~> 2.7.0'
-gem 'friendly_id'
 gem 'nokogiri', '~> 1.8.5'
 gem 'om'
-gem 'paperclip', '~> 6.1.0'
```
pgwillia: friendly_id and paperclip are only used by `profiles` as far as I can tell.

app/assets/stylesheets/ualib-custom.scss
```diff
@@ -258,7 +258,7 @@ h2{
 /* Custom styles - Footer
 -------------------------------------------------- */
 .footer{
-    background: asset_url("footer-bg.jpg") #1b4724 repeat-x;
+    background: image_url("footer-bg.jpg") #1b4724 repeat-x;
```
pgwillia: The images weren't showing up in development unless this.  Not sure about this change or why it wasn't needed elsewhere.

pgwillia:
I'm not sure if I got the images :100: Didn't see the public images used anywhere.  Also many `app/assets/images`. I didn't go into the stylesheets, but these should be cleaned up too.

Issue #1454, opened 2019-01-09
Databases that do not need EZProxy have proxy URL prepended/ones that need it do not


seanluyk:
**Describe the bug**
I recently noticed that there's a piece of code that's prepending the EZProxy URL (http://login.ezproxy.library.ualberta.ca/login?url=) to all database records, regardless of whether or not they need it. There's a related problem where databases that DO need the EXProxy URL prepended for off-campus access don't have it for some reason. What this means in practice is that a user accessing a database from Blacklight will be required to log in to EZProxy even if the resource isn't proxied and even if they're on campus, which is extremely annoying and vice versa, will not be prompted to login and will be directed to a vendor login page. I also noticed in the case of the NEOS Catalogue, that there's a pretty significant performance hit when accessing it in an EZProxy session. I noticed this because the NEOS catalogue had incorrectly been proxied in LibGuides (the SpringShare product where database assets are managed), so I turned this off, yet almost a month later, it's still being proxied. A search in the repository found the EZProxy URL, and I confirmed this with a number of examples. 

**To Reproduce**
1. Access the [NEOS Catalogue](https://library.ualberta.ca/catalog/11711067). "Click here for access"
2. Note that the EZProxy does not exist in the source (in LibApps). URL received on ingest should be something like http://ualweb.library.ualberta.ca//uhtbin/cgisirsi/x/0/0/57/49?user_id=WUAARCHIVE

**Expected behavior**
I see two possible ways of dealing with this:
1.  If possible, modify the database ingest script so it looks for the "Proxy Enabled" setting in LibGuides. If yes, prepend proxy. If no, don't! 
2. Modify the ingest script to pull in/use the LibGuides "Friendly URL" instead (e.g. http://guides.library.ualberta.ca/databases/neos), which redirects to the appropriate link (proxied or not proxied) 

**Additional context**
I need to touch base with CSU to confirm that this is no longer required and if there's a really good reason not to mess with anything. I think this is meant to be a failsafe, but I don't believe it's needed any longer

seanluyk:
@weiweishi this is high priority for discussion at the next grooming meeting/next spring as the problem is bigger than I first realized. I've updated the title and comment to reflect the fact that the EZProxy prefix isn't coming through when it should, which means users can't access our databases from off campus via Blacklight. I think the first step will be a spike into the database ingest code

pgwillia:
[...]

PR #1470, opened 2019-01-18
fix logic for when to add proxy to database url

pgwillia:
Add tests to catch regressions and rename the boolean test to be more 
meaningful and to miror the method name.

#1454

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -33,15 +33,15 @@ module CatalogHelper
   end
 
   def ual_electronic_access_url(location, url)
-    electronic_access_url(location.include?('Free Access'), url)
+    electronic_access_url(location.exclude?('Free Access'), url)
   end
 
   def database_electronic_access_url(document)
     electronic_access_url(document['enableproxy_tesim'].first == '1', document['url_tesim'].first)
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -37,7 +37,7 @@ def ual_electronic_access_url(location, url)
   end
 
   def database_electronic_access_url(document)
-    electronic_access_url(document['enableproxy_tesim'].first == '1', document['url_tesim'].first)
+    electronic_access_url(document['enableproxy_tesim'].first == '0', document['url_tesim'].first)
   end
 
   def electronic_access_url(test, url)
```
weiweishi: I wonder if we can change  "test" here to a more meaningful name of the variable? maybe `enable_proxy`?

pgwillia: I think I've improved this. What do you think?

seanluyk:
@weiweishi we're starting to get helpdesk tickets and emails from users about this issue. Is it possible to expedite its review, and getting it into prod soon?

app/helpers/catalog_helper.rb
weiweishi: This looks much better to me know

Issue #882, opened 2016-01-20
New Books List


redlibrarian:
ILS-team will produce list of catkeys and we can build in a static page.

redlibrarian:
This is list is being sent to me by email right now. We need to a) implement and b) automate this process.

redlibrarian:
I've built a proof of concept into the application, but we need to discuss how we want it to behave. I'll put it on the agenda for the next @ualbertalib/discovery-testing meeting (next Monday).

theLinkResolver:
@redlibrarian I checked and there is a MARC field for Date/time stamp (005) but it won't help us with this. I think the data you need will have to come from the item record:

![datecreated](https://cloud.githubusercontent.com/assets/14913579/14438018/3a837a68-ffe1-11e5-9c26-e3abecf0c231.jpg)

Items added to existing records (vs. newly-created records) are important to include, because another NEOS library may have something before we do, and we want to report our acquisition when it happens. However, we don't want to see single issues of serials (and also likely not volumes of multi-part monos) showing up in the lists as was mentioned in our meeting. 

It's like you need to look at the item creation date for a UA item but then disqualify the title if there is already another UA item (BUT you don't want to exclude when two items were created on the same day, e.g. a two-volume monograph).

redlibrarian:
All of a sudden the logic behind this has increased exponentially. Question for the @ualbertalib/discovery-testing - I've been assuming "new books" meant "new to UAL", but in a consortial environment, is this actually the case? People can request new books from anyway, so should the new book facet be built on "new to NEOS" logic instead?

redlibrarian:
I think this is unlikely to be implement in Release 3.0.

theLinkResolver:
I'd guess "new to UAL" is what people want.

redlibrarian:
Yes, apparently WAT decided that - but I think I might not have been there.

After conferring with @sjshores , the new plan is this: I won't be able to get the new books facet going for the May launch (the logic is too complicated for a quick fix), so I'm going to implement a basic static new books list with Jim's data for now, and then I'll look at the facet version in the summer.

redlibrarian:
[...]

PR #1471, opened 2019-01-21
Remove new books

murny:
This has been broken in production for who knows how long

image

Looks like this was the original ticket: https://github.com/ualbertalib/discovery/issues/882

So yes it never worked and was kinda just abandoned and left in production for years...

The error is a result of having a solr query that is too big (RSolr::Error::Http - 414 Request-URI Too Long), as we add 300 book ids to a solr query....

We can always revisit this (when the above ticket gets new life and this feature is deemed not out of scope) and do this properly (code will always be visible and versioned in git so can always revisit this later), but for now lets remove this


weiweishi:
@seanluyk I think we had a brief discussion in our last grooming meeting, or an ITS meeting - regarding the new book function. This PR looks good to me, as long as you don't have any objection to remove it (it doesn't work in production anyway, so if it's important, we'd need to develop better ways to handle it...)

seanluyk:
@weiweishi @murny this can be removed. There are much easier ways to accomplish this that we could look at instead (e.g. indexing data catalogued and then creating a static page with canned searches)

Issue #1113, opened 2017-05-08
[Discovery] Database license icons


redlibrarian:
**User story:** As a user, I want to see the icons for the different license/funding details on each database.

**Description**: The databases are catalogued in the [LibGuide platform](
https://guides.library.ualberta.ca/az.php), and one of the fields is for license/funding details. Each of the possible values for this field has an associated icon. When we index the databases in Discovery using the LibGuide API, we see the value of the **icons** field which we map to **icons_tesim** in Solr. Based on the value of **icons_tesim** we display the appropriate icon in Blacklight. There are four license/funding detail values: U of A Licensed, HKN Licensed, Lois Hole Digital Library, and Open Access/Free. Currently only the U of A Licensed icon has been [implemented](https://github.com/ualbertalib/discovery/blob/34db630ecd87b8ff7dc08cd71055b49769e28821/app/views/catalog/_index_default.html.erb#L33). Responsibility for cataloguing the databases, maintaining the icons in LibGuides, and managing the license/funding details lies with the Collections Strategies Unit.

**Done when**:
- [x] License/funding details data is parsed out of the LibGuides API.
- [x] Appropriate icons are displayed for each of the four conditions.

weiweishi:
A spike is needed to investigate if SpringShare API will return the icons field.

seanluyk:
Needs more discussion after the investigation. For example, could CORAL data be used instead?

seanluyk:
API documentation is behind a LibApps/guides login - anyone who wants access, please put in a helpdesk ticket. It looks like the following parameters using = expand could possibly give us the icon info:
 - icons
 - permitted_uses

pgwillia:
Can get 
```
"icons_tesim":["\n    \n      6665\n      165\n      ualib-logo.png\n      licensed for  U of A\n      https://www.library.ualberta.ca/about-us/policies/restricted-resource-info\n      11710698\n      6665\n      \n    \n    \n      15484\n      165\n      cou-user.png\n      User Registration Required\n      https://www.library.ualberta.ca/about-us/policies/conditions-of-use/registration-required\n      11710698\n      15484\n      \n    \n    \n      15485\n      165\n      cou-sign-in.png\n      CCID Login Required\n      https://www.library.ualberta.ca/about-us/policies/conditions-of-use/restricted\n      11710698\n      15485\n      \n    \n  "]
````
[...]

PR #1474, opened 2019-01-22
add database license icons

pgwillia:
Parse the mangled string stored in Solr to get the icon attributes. 
Display the icon on the results and show catalog pages.

screenshot from 2019-01-25 12-14-21
screenshot from 2019-01-25 12-17-44
screenshot from 2019-01-25 12-15-39

#1113


pgwillia:
@seanluyk What do you think about the dual use of the UAL shield? https://github.com/ualbertalib/discovery/issues/1113#issuecomment-456509823

app/views/catalog/_show_header_default.html.erb
```diff
@@ -17,4 +17,12 @@
       </div>
     <% end %>
   </div>
+  <p>
+    <strong>Conditions of Use:</strong>
```
weiweishi: Do we need to worry about the i18n in blacklight?

pgwillia:
@seanluyk Yes!  The descriptions are used as the alt value and images link out to their description.  Like this
```
<a href="https://www.library.ualberta.ca/about-us/policies/restricted-resource-info">
          <img alt="licensed for  U of A" src="https://libapps-ca.s3.amazonaws.com/sites/165/icons/6665/ualib-logo.png" height="46">
</a>
```

app/views/catalog/_show_header_default.html.erb
```diff
@@ -17,4 +17,18 @@
...
     <% end %>
   </div>
+  <div class='container'>
+    <div class='row row-no-gutters'>
+      <h2 class='col-md-3 h3 align-middle'><%= t('conditions_of_use') %>:</h2>
+      <ul class='col-md-9 align-middle list-inline'>
+      <% libguides_icons(document).each do |icon| -%>
+        <li class='list-inline-item'>
+        <%= link_to icon[:url] do %>
+          <%= image_tag(libguide_icon_image(icon), alt: icon[:description], height: '30') %>
```
murny: One small thing that could improve UX is we could add a `title` attr to the image tags? Be the same content as the `alt` attr

Why? This adds a nice tooltip when someone hovers over the icons giving some additional information of what the icons actually mean (aka hovering over the UofA shield icon will show a tooltip with `licensed for U of A`)
```suggestion
          <%= image_tag(libguide_icon_image(icon), alt: icon[:description], title: icon[:description], height: '30') %>
```

app/views/catalog/_show_header_default.html.erb
```diff
@@ -17,4 +17,18 @@
       </div>
     <% end %>
   </div>
+  <div class='container'>
```
murny: Shouldn't need to redefine a container div here? This whole partial gets rendered into a container already

pgwillia: When trying to get the description and the images to line up the Bootstrap documentation seemed to suggest that it was required.  I'm not sure it is either.  I'll check and remove if it's not.

murny: Yeah you need a container for rows/cols but we already have a container (every single thing in blacklight gets rendered inside a container that comes from the application layout: https://github.com/ualbertalib/discovery/blob/master/app/views/layouts/application.html.erb#L47)

Hope that makes sense can explain in more detail offline

PR #1349, opened 2018-10-24
Kule custom methods

piyapongch:
- Added Kule collection data
- Added id field on an item detail page

PR #1483, opened 2019-02-01
kule custom methods

pgwillia:
Co-Authored-By: Piyapong Charoenwattana <piyapong.charoenwattana@ualberta.ca>
screenshot from 2019-02-01 11-10-46
screenshot from 2019-02-01 11-10-01
screenshot from 2019-02-01 11-09-47

I was trying to see where #1349 was at but had a tough time seeing the relevant differences because of the history.  I've captured just the differences here and fixed the failing tests.

The kule marc fixture file is quite large.  Would be good to pair this down to a few illustrative examples.


pgwillia:
I moved `discovery` to `discovery_kule` on UAT because there are some customization and files there that I haven't seen elsewhere.

PR #1485, opened 2019-02-01
Remove unused partials

murny:
- Remove forsure unused partials (footer/home/navbar)
- Remove partials that havent changed from blacklight...just fallback to blacklights partials (until we can clean this stuff up properly) and most of these blacklight partials are done better!
- Remove some silly partials that are a few lines or smaller and move them to where they being rendered
- Fix deprecation  notice: 
image


app/views/catalog/_index_header_default.html.erb
```diff
@@ -18,7 +18,7 @@
       <% document.to_h['edition_tesim'] ? @edition = ", #{document.to_h['edition_tesim'].first}" : @edition = "" %>
       <% document.to_h['title_display'].nil? ? @combined_title = "Untitled document" : @combined_title = document.to_h['title_display'].first+@subtitle+@edition%>
 
-        <%= link_to_document document, :label=>@combined_title, :counter => counter %>
+        <%= link_to_document document, @combined_title, :counter => counter %>
```
murny: Simple fix to remove deprecation notice. Don't pass this in as a label hash, it's now just the second argument

app/views/catalog/_show_tools.html.erb
```diff
@@ -20,11 +20,15 @@
 
     <ul class="nav">
       <% if @document.export_formats.keys.include?( :refworks_marc_txt ) %>
-        <%= render "refworks" %>
```
murny: These partials just render a li with a link...just move this partials content here...

app/views/catalog/_sort_and_per_page.html.erb
```diff
@@ -1,9 +1,51 @@
 <div id="sortAndPerPage" class="clearfix">
   <%= render :partial => "paginate_compact", :object => @response %>
   <div class="search-widgets pull-right">
-    <%= render :partial => 'bookmark_widget' %>
```
murny: Just move all the partials here....now you get to see some similar things...

Noticed the common if statements? The first two partials have the exact same logic...now we can now refactor this and start combining html and throwing away things...magic!

Issue #1476, opened 2019-01-24
[Discovery] Finish cleaning up CMS project from Discovery


murny:
Issue https://github.com/ualbertalib/discovery/pull/1465 did a pretty job removing most of the CMS, but there is still a few chunks of code in discovery relating to the CMS that can be safely removed as well.

- [x] We got multiple CMS view layouts no longer being used (home/french/cms/cms_french)
- [x] We got lots and lots of CSS that only has to do with the now removed CMS
- [x] we also got lots of javascript relating to profile/cms pages (stupidtable, hoverintent, what is in ualib-home that we actually need? etc)
- [x] We got CMS stuff within our `bin/setup`, rake tasks (cmssetup.rake), in our README.md etc. that needs removal
- [x] I assume the CMS project is now using a different database? Let's drop these models from our schema/migrations
- [x] cleanup rubocop rules that had exceptions for CMS
- [x] cleanup ansible playbook

weiweishi:
@nmacgreg @nnunn Does CMS have its own database, or does it share the same DB with Discovery. There are DB related code we could remove from Discovery if CMS has its own.

nnunn:
library-cms has its own database so go ahead and remove that code!

murny:
#1493 pretty much completes this now. 

Created another ticket for a deeper cleanse of the CSS/JS here: https://github.com/ualbertalib/discovery/issues/1502

Also the checkbox for refactoring ansible will probably be a separate ticket (was on the priority list) and I'd assume this will be done here.

PR #1493, opened 2019-02-07
CMS cleanup

murny:
#1476


app/assets/javascripts/bootstrap.min.js
```diff
@@ -1,7 +0,0 @@
-/*!
```
murny: We never use bootstrap.min.js or bootstrap.min.css like this. It comes in from the bootstrap-sass gem and included via blacklight.

Get rid of these

app/assets/stylesheets/ualib-custom.scss
```diff
@@ -1136,82 +1136,6 @@ background: asset_url("search-choice-icon.png") #4C6F84 no-repeat 6px center;
 .title{
   margin: 25px 0 20px 0;
 }
-#news{
```
murny: There's probably lots of things like this "news" stuff in this file that should get purged.

See #1502

app/views/layouts/application.html.erb
```diff
@@ -17,9 +17,9 @@
     <%= opensearch_description_tag application_name, opensearch_catalog_url(format: 'xml') %>
     <%# favicon_link_tag asset_path('favicon.ico') %>
     <%= favicon_link_tag 'favicon.ico' %>
-    <%= stylesheet_link_tag 'application', 'https://fonts.googleapis.com/css?family=Open+Sans:400,700,400italic|Abel', 'sticky-footer', 'ualib-discovery', media: 'all' %>
```
murny: Should only be rendering one `application` css and js file. Let's use the asset pipeline properly here and have the application manifest contain the things we want like `ual-discovery` etc.

app/views/shared/_footer.html.erb
```diff
@@ -1,9 +1,3 @@
-<div class="push"></div>
```
murny: Let's fix this nonsense. Move the closing div tags where they are being used (in the layout)...and just delete the other two...

app/views/layouts/application.html.erb
pgwillia: For my own information...  What do you mean by 'application manifiest' is that this application layout file?

db/schema.rb
```diff
@@ -11,200 +11,56 @@
...
 
-ActiveRecord::Schema.define(version: 20151029171613) do
+ActiveRecord::Schema.define(version: 20190207211157) do
 
   create_table "bookmarks", force: :cascade do |t|
-    t.integer  "user_id",       null: false
-    t.string   "user_type"
-    t.string   "document_id"
-    t.string   "title"
+    t.integer  "user_id",       limit: 4,   null: false
```
pgwillia: Will adding limits to this schema have any impact on existing databases?

app/views/layouts/application.html.erb
murny: I'll explain this more today in the frontend knowledge share. 

But the manifest files will be the entry files for the the css/js (you can name them anything) typically the application.css/application.js files. 

So instead of including all the css/js files here in the layout and rendering them out with a stylesheet/javascript link tag. We instead should only be rendering the manifest files here...which contain all the css/js so this way sprockets and asset_pipeline can do some amazing stuff with it which saves performance, requests, and byte size for the end user.

db/schema.rb
murny: This schema file will never get ran in production/staging/etc. The schema file is just a way to setup the database without having to do go through all the migrations. Some projects might have 1000s of migration files, so running `rails db:create && rails db:migrate` might take a long long time. Think of this file as a cache?

When we run `rails db:setup` it will just create the database....load the schema from this file and run the seed data. Essentially skipping all the migrations (unless the version timestamp is out of date then it might just run the latest migrations)

So in production we would never do this. We would always just migrate the database with the latest migrations.

Hope this makes sense?

pgwillia: I feel like I've gotten that question before and was not confident in my answer.

PR #1480, opened 2019-01-29
Add subfields to Title and Contents for search and display

pgwillia:
This changes the subfields that are indexed for the title and content notes using the `rake ingest[symphony]` task.

#1121
`catalog/4220558`
Before:
screenshot from 2019-01-29 14-37-53
After:
screenshot from 2019-01-29 14-37-21
`catalog/2829082`
Before:
screenshot from 2019-01-29 14-41-48
After:
screenshot from 2019-01-29 14-41-33

#978
`catalog/910509`
Before:
screenshot from 2019-01-29 14-45-03
After:
screenshot from 2019-01-29 14-44-27

PR #1510, opened 2019-02-26
remove subtitle from display

pgwillia:
#1480 added a bunch of subfields to what is indexed for title, including 
subtitles. So this isn't necessary.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -49,11 +49,7 @@ class CatalogController < ApplicationController
 
     @additional_authors = @document['author_addl_t'] if @document['author_addl_t']
 
-    if @document['title_display']
-      @document['title_display'] = "#{@document['title_display'].first}: #{@document['subtitle_display'].first}" if @document['subtitle_display']
-    else
-      @document['title_display'] = 'Untitled document'
-    end
+    @document['title_display'] = 'Untitled document' unless @document['title_display']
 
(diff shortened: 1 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -49,11 +49,7 @@ def show
 
     @additional_authors = @document['author_addl_t'] if @document['author_addl_t']
 
-    if @document['title_display']
-      @document['title_display'] = "#{@document['title_display'].first}: #{@document['subtitle_display'].first}" if @document['subtitle_display']
```
pgwillia: There's a helper for header_title which helps with the .first part of this.

app/controllers/catalog_controller.rb
```diff
@@ -156,8 +152,6 @@ def show
     # config.add_show_field 'title_display', :label => 'Title'
     # config.add_show_field 'title_vern_display', :label => 'Title'
     config.add_show_field 'title_addl_t', label: 'Full/Alternate Title(s)'
-    # config.add_show_field 'subtitle_display', :label => 'Subtitle'
```
pgwillia: Removing irrelevant comments.

pgwillia:
> Do we also need to remove these:
> https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/symphony_index.properties#L33-L35
> or you plan to keep them in the index just in case?

#1517

Issue #1290, opened 2018-09-26
Publication year facet changes when adding a subject limiter


seanluyk:
**Describe the bug**
When clicking on a subject from a record view, the behaviour of the Publication year facet changes from its default. The expected date range slider is removed and is replaced with textual description of publication range and [View Distribution link](https://library.ualberta.ca/catalog/range_limit?q=%22Trees++United+States%22&range_end=2010&range_field=pub_date&range_start=1785&search_field=subject&view=brief) 

**To Reproduce**
1. Go to [this record](https://library.ualberta.ca/catalog/1236269)
2. Click on subject [trees>United States](https://library.ualberta.ca/catalog?q=%22Trees++United+States%22&search_field=subject&view=brief)
3. Scroll down to Publication Year facet
4. Expand facet

**Expected behavior**
The date range slider in the Publication Year facet should be consistently displayed, regardless of additional limiters being added

**Screenshots**
![screen shot 2018-09-26 at 16 25 30](https://user-images.githubusercontent.com/14880206/46112726-d1017400-c1a8-11e8-89f8-7ea917aeb415.png)

seanluyk:
@pgwillia another buggy one to fix if you have time

pgwillia:
Looks like the range facet is populated from an AJAX call.  It's not happening in the case when you come from the catalog page.
![screenshot from 2019-03-01 13-18-04](https://user-images.githubusercontent.com/1220762/53664104-d44fce80-3c24-11e9-9d18-dd3309dbb427.png)
![screenshot from 2019-03-01 13-17-22](https://user-images.githubusercontent.com/1220762/53664105-d44fce80-3c24-11e9-8223-aa51b279fb28.png)
I think that turbolinks may be responsible.  I'm pursuing that angle.

seanluyk:
Fixed in #1520, closing

PR #1520, opened 2019-03-01
turn off turbolinks on the show page for range_limit

pgwillia:
screenshot from 2019-03-01 13-18-04

The range_limit XHR request is what gives the context to popluate the
range facet. While turbolinks helps to improve the perceived
performance, it prevents the XHR range_limit request from being loaded.
Links with Turbolinks disabled will be handled normally by the browser.

Also refactor links on show page to use rails helper link_to

#1290

### DIFF
--- a/app/views/catalog/_show_default.html.erb
+++ b/app/views/catalog/_show_default.html.erb
@@ -6,41 +6,36 @@
       <dt class="blacklight-<%= solr_fname.parameterize %>">
         <%= render_document_show_field_label document, field: solr_fname %>
       </dt>
+      <% value = render_document_show_field_value document, field: solr_fname %>
       <% if @linked_fields.include? solr_fname %>
         <dd class="blacklight-<%= solr_fname.parameterize %>">
-          <a href="/catalog?q=&quot;<%= render_document_show_field_value(document, field: solr_fname).tr(' ', '+') %>&quot;">
-            <%= render_document_show_field_value document, field: solr_fname %>
-          </a>
+          <%= link_to value, catalog_index_path(q: "\"#{value}\"") %>
         </dd>
       <% elsif solr_fname == "title_series_t" %>
         <dd class="blacklight-<%= solr_fname.parameterize %>">
-          <a href="/<%= params[:controller] %>?title=&quot;<%= render_document_show_field_value(document, field: solr_fname).split(';').first.strip.tr(' ', '+') %>&quot;&search_field=advanced">
-            <%= render_document_show_field_value document, field: solr_fname %>
-          </a>
+          <%= link_to value, catalog_index_path(title: "\"#{value}\"", search_field: 'advanced') %>
         </dd>
       <% elsif solr_fname == "author_display" %>
         <dd class="blacklight-<%= solr_fname.parameterize %>">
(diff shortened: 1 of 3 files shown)


app/views/catalog/_show_default.html.erb
```diff
@@ -6,41 +6,36 @@
       <dt class="blacklight-<%= solr_fname.parameterize %>">
         <%= render_document_show_field_label document, field: solr_fname %>
       </dt>
+      <% value = render_document_show_field_value document, field: solr_fname %>
```
weiweishi: like how you capture this into a variable - make it much easier to read.

app/views/catalog/_show_default.html.erb
```diff
@@ -6,41 +6,36 @@
...
       <% if @linked_fields.include? solr_fname %>
         <dd class="blacklight-<%= solr_fname.parameterize %>">
-          <a href="/catalog?q=&quot;<%= render_document_show_field_value(document, field: solr_fname).tr(' ', '+') %>&quot;">
-            <%= render_document_show_field_value document, field: solr_fname %>
-          </a>
+          <%= link_to value, catalog_index_path(q: "\"#{value}\""), data: { turbolinks: false } %>
```
weiweishi: And the use the proper `link_to` instead of hardcoded urls :+1:

app/views/catalog/_show_default.html.erb
weiweishi: there were lots of string manipulation that replace space with "+", why we won't need them now? wouldn't `link_to` use `%20` for spaces if they exist?

weiweishi:
LGTM, just one small question.

app/views/catalog/_show_default.html.erb
pgwillia: I was trying find a definitive source but failed. This was the closest: [Segments that are passed into URL helpers are now automatically escaped](https://guides.rubyonrails.org/4_2_release_notes.html). It appeared to me that the Rails url helpers are smart about properly escaping -- one of the advantages of using the helpers rather than trying to do it ourselves. 

```
8] pry(main)> app.catalog_index_path(q: ' +')
=> "/catalog?q=+%2B"
```
Notice that the space ' ' becomes '+'.  Also notice that '+' becomes '%2B'.

weiweishi: okay, thanks for the investigation - tests seem to be fine too, so let's just trust the smart rails URL helpers and will keep an eye for any potential issue.

seanluyk:
@pgwillia @weiweishi [3.0.100](https://github.com/ualbertalib/discovery/releases/latest) looks good except for this fix. The range slider has been replaced by a long list of publication years:
image

weiweishi:
Interesting how this works in UAT:
http://uat.library.ualberta.ca:3002/?q=test
Will investigate.


*Weiwei Shi*Digital Initiative Applications Librarian

2-10L Cameron Library, University of Alberta
780-492-7802  |  weiwei.shi@ualberta.ca
*"The University of Alberta respectfully acknowledges that we are situated
on Treaty 6 territory, traditional lands of First Nations and Métis
people."*

seanluyk:
Thanks @weiweishi something else funny there too, but possibly a css/assets issue? There's no histogram appearing in the range limiter in UAT

weiweishi:
With a redeployment of UAT, I can replicate the same behaviour as on search-test.

Issue #1432, opened 2018-12-14
Upgrade Ruby to Latest


murny:
We talked about this for a few months now but nothing has been done yet about it. Here's an issue to track this.

We are using Ruby 2.1.5 in production. This version was released over 4 years ago! It is now unsupported and has been End of Life well over a year now! What does this mean? This means our Ruby version receives no more updates to it. This should be a big security concern since this version receives no security updates to it (which there has been quite a few present since this version has been EOL: https://www.ruby-lang.org/en/security/). There is also some major performance gains too from updating.

This is also prevented us from using the latest of other software like bootstrap and sass: https://github.com/ualbertalib/discovery/pull/1431#discussion_r241870723

What needs to be done? Nothing from the development side. Discovery works fine with the latest ruby version. We need to update the ansible playbook and bump this version.

(On a similar note we are using very old passenger and apache versions too which we tell every end user about this via our response headers (noted here as we do the same thing in jupiter: https://github.com/ualbertalib/jupiter/issues/703)

pgwillia:
Latest is 2.6.  While current code base works well with 2.5 there appears to be a bit of work to get it going with [2.6](https://travis-ci.org/ualbertalib/discovery/builds/476933977).

murny:
Agreed. When I created this issue 2.5 was the latest. So I would stick with `2.5.X`. 

Version 2.6  came out on Christmas so probably not quite ready to move to this version yet.

PR #1528, opened 2019-03-07
update ruby 2.5.3

pgwillia:
- remove old ruby dependency
- fixed rubocop nags
- unlock and update rubygems
- reintroduce erblint and fix nags
- address some deprecation warnings.

#1432

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,7 +53,7 @@ group :development do
   gem 'better_errors', '>= 2.3.0'
   gem 'binding_of_caller'
 
-  gem 'erb_lint', '0.0.28', require: false
+  gem 'erb_lint', '~> 0.0.28', require: false
 
   gem 'letter_opener'
 
--- a/Gemfile
+++ b/Gemfile
@@ -9,12 +9,9 @@ gem 'mysql2', '~> 0.4.10'
 
 # Assets (CSS/JS) stuff
 gem 'jquery-rails'
-gem 'sass-rails', '~> 5.0.1'
+gem 'sassc-rails', '~> 2.1'
 gem 'turbolinks', '5.2.0'
 gem 'uglifier', '>= 1.3.0'
-
-# FIXME: Since we stuck using EOL Ruby 2.1.5, need to lock this down...
-gem 'bootstrap-sass', '3.3.7'
 
 # blacklight stuff
 gem 'blacklight', '5.15.0'
@@ -37,8 +34,6 @@ gem 'rollbar'
 
 group :test, :development do
   gem 'sdoc', require: false
-  # FIXME: Since we stuck using EOL Ruby 2.1.5, need to lock this down...
-  gem 'rdoc', '5.1.0'
 
   gem 'factory_bot'
   gem 'rspec'
@@ -46,9 +41,8 @@ group :test, :development do
   gem 'rspec-solr'
 
   gem 'pry'
+  gem 'pry-byebug'
   gem 'pry-rails'
-  # FIXME: Since we stuck using EOL Ruby 2.1.5, need to lock this down...
-  gem 'pry-byebug', '3.4.3'
 
   gem 'rubocop', '~> 0.56.0', require: false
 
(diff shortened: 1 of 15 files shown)


Gemfile
```diff
@@ -9,13 +9,10 @@ gem 'mysql2', '~> 0.4.10'
 
 # Assets (CSS/JS) stuff
 gem 'jquery-rails'
-gem 'sass-rails', '~> 5.0.1'
+gem 'sassc-rails', '~> 2.1'
 gem 'turbolinks', '5.2.0'
 gem 'uglifier', '>= 1.3.0'
 
-# FIXME: Since we stuck using EOL Ruby 2.1.5, need to lock this down...
-gem 'bootstrap-sass', '3.3.7'
```
weiweishi: Do we need to add bootstrap 4 to the dependency? now we've removed bootstrap-sass?

Gemfile
```diff
@@ -59,6 +53,8 @@ group :development do
   gem 'better_errors', '>= 2.3.0'
   gem 'binding_of_caller'
 
+  gem 'erb_lint', '0.0.28', require: false
```
weiweishi: we still need to lock down the version for this gem?

Gemfile
```diff
@@ -70,10 +66,7 @@ end
 group :test do
   gem 'simplecov', require: false
 
-  # FIXME: Since we stuck using EOL Ruby 2.1.5, need to lock this down...
-  gem 'capybara', '2.18.0'
-  gem 'xpath', '2.1.0'
```
weiweishi: no longer need the xpath gem?

app/models/solr_document.rb
```diff
@@ -20,7 +20,7 @@ class SolrDocument
   SolrDocument.use_extension(Blacklight::Document::Email)
 
   # SMS uses the semantic field mappings below to generate the body of an SMS email.
-  SolrDocument.use_extension(Blacklight::Solr::Document::Sms)
+  SolrDocument.use_extension(Blacklight::Document::Sms)
```
weiweishi: Would you mind explain this a little bit? Not sure if I know why we had to call Blacklight::Solr before for this one in particular, but not right now with a Ruby upgrade. Blacklight gem doesn't seem to have changed?

Gemfile
pgwillia: If you inspect the Gemfile.lock you'll see that bootstrap-sass (~> 3.2) is a dependency of blacklight.  We don't need to explicitly pin this anymore, blacklight will take care of it.

Gemfile
pgwillia: Similar to above.  xpath is a dependency of capybara.  We don't need to explicitly pin anymore, capybara will take care of it.

app/models/solr_document.rb
pgwillia: Your right.  Not related to the ruby upgrade.  It's just a deprecation warning I noticed and thought to clear up.

Gemfile
pgwillia: This is how Shane had introduced it: https://github.com/ualbertalib/discovery/commit/b1e80f8c83ca6a258b67607df2931719fdeb4a45#diff-8b7db4d5cc4b8f6dc8feb7030baa2478  I don't think it's a bad thing to pin it.  What's your concern?

pgwillia: This is the latest version. [Rubygems](https://rubygems.org/gems/erb_lint) recommends `gem 'erb_lint', '~> 0.0.28'`

weiweishi: My concern is locking down on a specific version of gem and not getting hotfixes/minor upgrades.  This is the linter and won't impact the application itself, so I'm less concerned. But it would be nice to get the updates if they are available, and we can. I'd feel more comfortable with `gem 'erb_lint', '~> 0.0.28'`, but I guess linter is different than regular gems? there might be rules/syntaxes changes in between versions and would create additional work for us.

pgwillia: If your concerned about getting updates, I think dependabot has us covered.

Using Gentile.lock will pin this until we purposefully change it also.

pgwillia:
```suggestion
  gem 'erb_lint', '~> 0.0.28', require: false
```

pgwillia:
Added the erblint commit.

PR #1536, opened 2019-03-12
Pin blacklight range limit

pgwillia:
To address the issue noted here https://github.com/ualbertalib/discovery/pull/1520#issuecomment-471663175
Screenshot from 2019-03-12 10-54-56

This occurred because the version of blacklight_range_limit was bumped (with the negative consequences noted) https://github.com/ualbertalib/discovery/pull/1528/commits/381faff26a6550e40d1f363485ea566acf6610f7#diff-e79a60dc6b85309ae70a6ea8261eaf95R82

While I'm here, I'll also add a config change to the Solr docker in the lightweight config.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem 'blacklight', '5.15.0'
 gem 'blacklight-marc', '~> 5.10.0'
 gem 'blacklight_advanced_search'
 gem 'blacklight_google_analytics'
-gem 'blacklight_range_limit'
+gem 'blacklight_range_limit', '~> 5.2.0'
 
 # Authentication
 gem 'devise'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -18,7 +18,7 @@ gem 'blacklight', '5.15.0'
 gem 'blacklight-marc', '~> 5.10.0'
 gem 'blacklight_advanced_search'
 gem 'blacklight_google_analytics'
-gem 'blacklight_range_limit'
+gem 'blacklight_range_limit', '~> 5.2.0'
```
weiweishi: good catch.

pgwillia:
> Good catch @pgwillia. I wonder why I didn't see this change in the previous PR. Do you think when we fork blacklight, should we lock all the blacklight related gem versions at that point?

Here is Shane's take
> Repeat similar steps for the Blacklight gems like Blacklight advanced search, Blacklight marc, Blacklight google analytics, Blacklight range limit all of which are locked to a version and all of which provide hardly any functionality and some of them we probably don't really care about at all or getting 2 cents of value out of them?

Makes sense to me. The Gemfile.lock has them locked.  We could add versions to the Gemfile as well so that mistakes like mine are less likely to happen.

Issue #1549, opened 2019-03-18
Bookmarks Broken


seanluyk:
**Describe the bug**
Navigating to saved bookmarks leads to a 404 error

**To Reproduce**
1. Do a search
2. Add some items to bookmarks
3. Go to bookmarks https://search.library.ualberta.ca/bookmarks

**Additional context**
Bookmarks page loads before items are added, so the error is related to adding items

seanluyk:
@weiweishi this is the issue I mentioned this AM that has some helpdesk tickets

weiweishi:
I think this issue is related to the part where it renders the refworks link:
https://rollbar.com/ualbertalib/discovery/items/97/
where the error is thrown
```
<%= link_to t('blacklight.tools.refworks'), refworks_export_url(url: bookmarks_export_url(:refworks_marc_txt, params_for_search)), class: 'btn btn-default', id: 'refworksLink', target: '_blank' %>
```

the error message is related to key length, and it's probably because of this:
https://github.com/rails/rails/pull/25192?

seanluyk:
@weiweishi good sleuthing, hopefully, this can be fixed in this release? Wondering if it's related to the RefWorks export issue too?

pgwillia:
More specifically this is encrypt_user_id https://github.com/projectblacklight/blacklight/blob/v5.15.0/lib/blacklight/token_based_user.rb#L37-L41
```
# Used for #export action with encrypted user_id, available
# as a helper method for views.
def encrypt_user_id(user_id)
  message_encryptor.encrypt_and_sign([user_id, Time.zone.now])
end

##
# This method provides Rails 3 compatibility to our message encryptor.
# When we drop support for Rails 3, we can just use the AS::KeyGenerator
# directly instead of this helper.
def export_secret_token salt
  OpenSSL::PKCS5.pbkdf2_hmac_sha1(Blacklight.secret_key, salt, 1000, 64)
end

def message_encryptor
  derived_secret = export_secret_token("encrypted user session key")
  ActiveSupport::MessageEncryptor.new(derived_secret)
end
```

seanluyk:
Fixed in production, but there's a new issue I'll add to a separate ticket

PR #1552, opened 2019-03-20
temporarily remove refworks export to fix issue in bookmark

weiweishi:
This is to temporarily remove refworks export so that the error message identified in #1549 won't break the bookmark feature.


weiweishi:
Working on adding the feature test now and will commit as a separate PR so it doesn't block the deployment of this hot fix.

pbinkley:
Minor point - we could throw a timestamp or even the link to this PR into the comment explaining the reason for the change, so that if this hangs around longer than expected a future dev will have an easy way to find the context. Not a blocker but worth thinking about for this kind of change.

PR #1559, opened 2019-03-22
Fix the 500 error message after email is sent from bookmarks page

weiweishi:
Bookmarks should have its own ```action_success_redirect_path```
https://github.com/projectblacklight/blacklight/blob/master/app/controllers/concerns/blacklight/bookmarks.rb#L31
But since we overwrite this:
https://github.com/ualbertalib/discovery/blob/master/lib/blacklight/catalog.rb#L113

The page redirects to /catalog instead of /bookmarks. 

This brings back bookmarks_controller to bring back this behavior for bookmarks. also updated the view with proper css.


app/views/bookmarks/_tools.html.erb
```diff
@@ -1,9 +1,10 @@
-<ul class="<%= controller_name %>Tools nav nav-pills">
+  <ul class="<%= controller_name %>Tools nav nav-pills">
   <%= render_show_doc_actions document_list, document: nil, document_list: @document_list, url_opts: sanitize_search_params(params) do |config, inner| %>
-    <li>
-      <%= inner.insert(3, 'class="btn btn-default"') %>
+    <li class="nav-item">
```
pgwillia: What's happening in this change?

weiweishi: I brought the lastest code from Blacklight, as this is the native style from bootstrap to handle individual navigation pill.

pgwillia: I'm not sure that [navigation](https://getbootstrap.com/docs/4.0/components/navs/) is really what this is. There isn't a state where the email nav pill will be active.  I think this is a [button](https://getbootstrap.com/docs/4.0/components/buttons/) that will take you away from this page to do something.

Before
Screenshot from 2019-03-22 10-54-04
After
Screenshot from 2019-03-22 10-54-27

weiweishi:
creating tests has been proved to be difficult, created a separate issue so that we will continue working on the tests while being able to deploy this fix to address the issue in production.

app/views/bookmarks/_tools.html.erb
weiweishi: And that's what it was - on any regular catalog record email function, the email dialogue is a popup window, and on the bookmarks page, it was a link to a static page at bookmarks/email. I made the change to make them consistent. But if you have concerns over this, we can move forward without it.

Issue #1545, opened 2019-03-14
Relabel "Other Physical Details" to "Physical Details"


seanluyk:
**Describe the bug**
The current label for information in the 300 field (Other Physical Details) is misleading as they are the only physical details! In an effort to improve the labeling on the UI, this is a quick win

**Expected behavior**
Relabel "Other Physical Details" to "Physical Details" 

**Additional context**
 - Discussed a long time ago in #783, which is how we came up with the current label

seanluyk:
Looking good!

PR #1578, opened 2019-03-26
Relabel "Other Physical Details" to "Physical Details"

ConnorSheremeta:
#1545


pgwillia:
This looks great!

Before: Screenshot from 2019-03-26 13-11-04
After: Screenshot from 2019-03-26 13-10-44

One of the things we're all trying to get into the habit of doing is to add a note in the CHANGELOG.md for each change. What do you think of adding a 'Changed' section to 'Unreleased' with something like your commit message?

PR #1583, opened 2019-03-28
Add Google Analytics for search.library.ualberta.ca

ConnorSheremeta:
I see that secrets.yml has the comment "In Production, this file is now managed by ansible, starting Jan 29, 2016, to simplify a secure deployment.  Please know that changes you make to this file will be overwritten on a Production Blacklight server!".

I'm assuming that the statement "(requires new secret)" inside of the changelog is enough to ensure that whoever is deploying will add the new secret.


ConnorSheremeta:
Looking at it again it does look like it would be better to match Jupiter's ga implementation, pushed a new commit.

PR #1585, opened 2019-03-29
Add Kule Collection Ingest Task

piyapongch:


Gemfile
```diff
@@ -61,6 +61,10 @@ group :development do
   gem 'listen', '~> 3.0'
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
+
+  # Debug
```
weiweishi: @piyapongch do you want to add some documentation maybe on the di_internal regarding how you set up you IDE environment, and use these debugging tools, and provide a link the documentation here?

app/helpers/holdings_helper.rb
```diff
@@ -61,4 +61,23 @@ def request_microform?(item)
...
     SYMPHONY_LIBRARY_LOCATIONS[library_code.downcase.delete('_').to_sym]
   end
+
+  def kule_holdings(document, holdings)
+    document[:callnumber_tesim].each do |values|
+      callnumber = values.split('|')
+      item = {}
+      item[:callnumber] = callnumber[0]
+      item[:status] = case callnumber[1]
+                      when 'NO_LOAN' then 'Read On Site'
```
weiweishi: I will follow up with @sfarnel because I don't think we should develop to fix the data issue. this should be addressed in the data level.

app/helpers/holdings_helper.rb
```diff
@@ -61,4 +61,23 @@ def request_microform?(item)
   def library_location(library_code)
     SYMPHONY_LIBRARY_LOCATIONS[library_code.downcase.delete('_').to_sym]
   end
+
+  def kule_holdings(document, holdings)
+    document[:callnumber_tesim].each do |values|
```
weiweishi: so this is the solr field that will have both callnumber and status? I wonder if we can rename it to call_number_status to avoid confusion?

config/SolrMarc/index_scripts/batch_number.bsh
```diff
@@ -0,0 +1,22 @@
+import org.marc4j.marc.Record;
```
weiweishi: I don't quite understand what this script is doing? Maybe I overlooked something?

config/SolrMarc/kule_index.properties
```diff
@@ -147,8 +119,14 @@ gmd_tesim = script(gmd.bsh), getGMD
...
 summary_holdings_tesim = 8563
-isbn_tesim = 020a, (pattern_map.isbn_clean)
 issn_tesim = 022a, first
 uniform_title_tesim = 240a
 performers_tesim = 511a
 publisher_number_tesim = 028a
+accession_number_tesim = 024a
+
+# Holding information
+callnumber_tesim = custom, getAllSubfields(926[at], "|")
```
weiweishi: Same as above? This field contains both call number and status. Probably should rename for clarity.

db/schema.rb
```diff
@@ -25,6 +25,150 @@
 
   add_index "bookmarks", ["user_id"], name: "index_bookmarks_on_user_id", using: :btree
 
+  create_table "comfy_cms_blocks", force: :cascade do |t|
```
weiweishi: you may want to rebase your branch with master. These has been removed from master branch already, and should not be part of your PR.

spec/lib/tasks/ingest_rake_spec.rb
```diff
@@ -8,7 +8,7 @@
     ENV['SOLR_INGEST_URL'] = Blacklight.connection_config[:url]
     expect { Rake::Task['ingest_info'].invoke }.to output(
       'target is set from environment variable SOLR_INGEST_URL=http://localhost:8983/solr/discovery-test'\
-      "\nSolr collection contains 10024 results.\n"
+      "\nSolr collection contains 10124 results.\n"
```
weiweishi: I wonder if we can add some specific tests for this particular kule ingest rake task?

weiweishi:
This looks good to me. Thanks for your work on this. I've left some comments on your PR.

PR #1586, opened 2019-03-29
test for an item with SFX lookup

pgwillia:
Trying to fill out the tests for different cases.  Also illustrates what a SFX response looks like in VCR cassette.


spec/fixtures/vcr_cassettes/sfx_result.yml
```diff
@@ -0,0 +1,108 @@
+---
```
weiweishi: I'd love to learn how do you record these vcr testing data.

Issue #1544, opened 2019-03-14
Create Responsibility field for record view, to appear below title


seanluyk:
**Is your feature request related to a problem? Please describe.**
To free up space/improve the display of 245$c, statement of responsibility, it would be beneficial to display this in a separate responsibility field outside of the title. Here's an example of this approach: https://searchworks.stanford.edu/view/11056542

**Describe the solution you'd like**
Display 245$c in a Responsibility field, below the title/subtitle

**Additional context**
 - Relates to #1521

seanluyk:
Just checking 3.0.106 and not seeing this in test FYI @weiweishi @ConnorSheremeta

weiweishi:
looking at the code, I think a reindex is needed for this to take effect.


*Weiwei Shi*Digital Initiative Applications Librarian

2-10L Cameron Library, University of Alberta
780-492-7802  |  weiwei.shi@ualberta.ca
*"The University of Alberta respectfully acknowledges that we are situated
on Treaty 6 territory, traditional lands of First Nations and Métis
people."*

seanluyk:
@weiweishi right, that makes sense

PR #1590, opened 2019-04-02
Create Responsibility field for record view, to appear below title

ConnorSheremeta:
#1544
I arbitrarily chose to have the 'Responsibility' field between 'Author' fields and 'Format', let me know if it would be better suited elsewhere.

 
#1544SS2


CHANGELOG.md
```diff
@@ -30,9 +30,9 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 ### Changed
 - refactor summary_holdings to reflect [change to NodeSet#each in nokogiri](https://github.com/sparklemotion/nokogiri/issues/1822) [PR#1533](https://github.com/ualbertalib/discovery/pull/1533)
 - record vcr cassettes for symphony service tests [PR#1533](https://github.com/ualbertalib/discovery/pull/1533)
+- Created Responsibility field for record view appearing below title [#1544](https://github.com/ualbertalib/discovery/issues/1544)
```
weiweishi: @ConnorSheremeta You'd want to add new changes/fixes under the ## [unreleased] section, which will become the section for the next release when we are ready to tag.

app/controllers/catalog_controller.rb
```diff
@@ -159,6 +159,7 @@ def show
     config.add_show_field 'author_display', label: 'Author'
     config.add_show_field 'author_addl_t', label: 'Additional authors/performers'
     config.add_show_field 'author_vern_display', label: 'Author'
+    config.add_show_field 'responsibility_display', label: 'Responsibility'
```
weiweishi: I wonder if we'd need to add this field to right under the 'title' field (same as how Stanford has done: https://searchworks.stanford.edu/view/11056542 @seanluyk what do you think?

seanluyk: @weiweishi @ConnorSheremeta I think we should follow Stanford's approach and up responsibility right under the main title

Issue #1589, opened 2019-04-02
Bookmarks email completion behaviour


seanluyk:
**Describe the bug**
Submitting the email form from bookmarks leads to a 404 error page

**To Reproduce**
1. Add some items to bookmarks
2. Click on email
3. Complete and submit form

**Expected behavior**
User should be returned to bookmarks page and email sent message should appear

**Additional context**
The email actually does get sent

weiweishi:
This is done and closing. Waiting for deployment to production.

Issue #1580, opened 2019-03-26
Add proper tests for bookmark controller


weiweishi:
We will need to add proper functional tests for the bookmark controller, for testing the basic function of creating, deleting and emailing bookmarks. Attempts have been made as part of the #1552 and https://github.com/ualbertalib/discovery/pull/1562. Because of the time sensitivity of resolving this issue #1552 , we will address this dev work in a separate PR.

weiweishi:
This is completed.

PR #1593, opened 2019-04-02
remove turbolinks to get consistant behaviour out of blacklight modals

pgwillia:
turbolinks seems to have caused some of our recent javascript related troubles.  I noticed that Sam pined the version to 2.5.x to solve some of his troubles https://github.com/ualbertalib/discovery/commit/caadb47803288335c80a34109e4e9c4e1e22fba3

This solution does need some more investigation about impact to performance or if with a little more TLC that we can upgrade to turbolinks 5.

Also add controller and feature tests for the bookmarks email feature.

remove the previous attempt to fix the email redirect error

fix https error in search_facet_url
Screenshot from 2019-04-02 15-32-37

#1589 #1580


CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 - Created Responsibility field for record view appearing below title [#1544](https://github.com/ualbertalib/discovery/issues/1544)
 - made google analytics token configurable through an environment variable (requires new secret) [#1545](https://github.com/ualbertalib/discovery/issues/1545)
 
+### Added
+- feature and integration tests for emailing bookmarks [#1580](https://github.com/ualbertalib/discovery/issues/1580)
+
+### Fixed
+- clicking 'more' on the Call Number facet [PR#1593](https://github.com/ualbertalib/discovery/pull/1593)
```
pgwillia: To reproduce:
1. Search for anything in the root path
2. Click on the "Call Number" facet on the left hand side
3. Click on 'more' to view more facets
4. Observe that nothing happens.

Oddly in search-test.library.ualberta.ca this works to produce the modal with the ability to browse more facets.  In development even mousing over that link will generate an error.

pgwillia: Should write a test to catch regresssions

pgwillia:
The fix for the facet was to remove HTTPS so I don't think this is asset compilation anymore.

weiweishi:
awesome

app/views/catalog/_facet_limit.html.erb
```diff
@@ -5,7 +5,7 @@
   <% unless paginator.last_page? || params[:action] == "facet" %>
     <li class="more_facets_link">
       <%= link_to t('more', scope: 'blacklight.search.facets', default: :more_html, solr_field: facet_field.label),
-                  search_facet_url(id: solr_field, protocol: 'https'), class: 'more_facets_link' %>
```
pgwillia: https://github.com/ualbertalib/discovery/blob/46a9fec812bc336f373b65bb68982b0ceca49c74/config/environments/production.rb#L1

Issue #1517, opened 2019-02-28
Remove subtitle infavour of having this all in the title field


pgwillia:
If we're comfortable with removing subtitle in favour for a more inclusive title field we can clean up the rest of the ingest code https://github.com/ualbertalib/discovery/blob/939731db09e2db879f01b92b8f9185307458e903/config/SolrMarc/symphony_index.properties#L33-L35

and solr config (subtitle_unstem_search, subtitle_t, subtitle_display, subtitle_vern_display) https://github.com/ualbertalib/blacklight_solr_conf/blob/bc89d514ca857214c57bc2cb023e0d6dfe4a79c6/solr/blacklight-core/conf/solrconfig.xml#L398

This is related to https://github.com/ualbertalib/discovery/issues/1309 and https://github.com/ualbertalib/discovery/issues/978.

seanluyk:
@pgwillia I think this will be fine, the only reason it wouldn't be is if we ever would want to separate this info for display purposes

Issue #1378, opened 2018-11-06
Deployment of release 3.0.92


pgwillia:
- [x] removed `therubyracer` gem.  Will need to ensure that `nodejs` is installed before assets can be built.

pgwillia:
We're removing therubyracer which previously was a dependency in building the assets.  This practice is no longer recommended https://devcenter.heroku.com/articles/rails-asset-pipeline#therubyracer rails/rails#29285. Instead, we will need to install nodejs. Here's a reference of how that's done in jupiter.

The same command bundle exec rake assets:precompile is run.  One outcome of this is that it might not take as long for this step.

PR #1594, opened 2019-04-04
Remove subtitle infavour of having this all in the title field

ConnorSheremeta:
#1517


ConnorSheremeta:
I believe so, what would be the best way to convey that to whoever will deploy?

pgwillia:
> I believe so, what would be the best way to convey that to whoever will deploy?

I think if we include it in the CHANGELOGs, then it can be noted in the release notes, and then it can be noted when we notify Neil by email of the release that contains the change.  Another thing we started to do, but it wasn't working well because we didn't go back to check for it was to create an issue for deployments ([like this](https://github.com/ualbertalib/discovery/issues/1378)) to make notes about deployment.

PR #1595, opened 2019-04-05
Add danger that enforces requirements for PR to Discovery

ConnorSheremeta:
- Added danger gem and danger file to the project


Dangerfile
```diff
@@ -0,0 +1,29 @@
...
+# Ensure a clean commits history
+if git.commits.any? { |c| c.message =~ /^Merge branch '#{github.branch_for_base}'/ }
+  warn 'Please rebase to get rid of the merge commits in this PR'
+end
+
+# Warn when there is a big PR
+warn('This PR is too big! Consider breaking it down into smaller PRs.') if git.lines_of_code > 500
+
+# Let people say that this isn't worth a CHANGELOG entry in the PR if they choose
+declared_trivial = (github.pr_title + github.pr_body).include?("#trivial") || !has_app_changes
```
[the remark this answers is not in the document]
pgwillia: Just thinking about @dependabot PRs.  They just change Gemfile and Gemfile.lock so would be in the `!has_app_changes` category.

Dangerfile
```diff
@@ -0,0 +1,29 @@
...
+
+if !git.modified_files.include?("CHANGELOG.md") && !declared_trivial
+  fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/danger/danger/blob/master/CHANGELOG.md).", sticky: false)
```
pgwillia:
```suggestion
  fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/ualbertalib/discovery/blob/master/CHANGELOG.md).", sticky: false)
```

Dangerfile
```diff
@@ -0,0 +1,29 @@
+# Make sure non-trivial amounts of code changes come with corresponding tests
+has_app_changes = !git.modified_files.grep(/lib/).empty?
```
pgwillia: The app code lives mostly in the `app` directory but also the `lib` directory
```suggestion
has_app_changes = !git.modified_files.grep(/lib\|app/).empty?
```

Gemfile.lock
```diff
@@ -89,29 +89,51 @@ GEM
     bootstrap_form (2.7.0)
     brakeman (4.5.0)
     builder (3.2.3)
-    byebug (11.0.0)
-    capybara (2.18.0)
+    byebug (11.0.1)
+    capybara (3.16.1)
```
pgwillia: There are some gem bumps that I wouldn't have expected. Were these intended?

ConnorSheremeta:
Ignore the warning about rebasing to get rid of merge commits as that can be done when merging (which I haven't been doing - oops!) Maybe it would be better to get rid of that check all together?

Issue #1592, opened 2019-04-02
Add new languages location


seanluyk:
There's a new location in Symphony that will need to be added to [statuses.yml](https://github.com/ualbertalib/discovery/blob/master/config/statuses.yml)  so holdings don't appear blank:

languages: "languages"

ConnorSheremeta:
Added issue for NEOS discovery [here](https://github.com/ualbertalib/NEOSDiscovery/issues/264)

weiweishi:
This is done. Closing

PR #1604, opened 2019-04-10
Add new languages location

ConnorSheremeta:
-Added missing location 'languages' into statuses.yml
#1592


ConnorSheremeta:
Added issue for NEOS discovery [here](https://github.com/ualbertalib/NEOSDiscovery/issues/264)

Issue #1635, opened 2019-05-06
Add feature tests for display of google book jackets


ConnorSheremeta:
Display of Google book jackets
From epic #1600

PR #1637, opened 2019-05-07
Feature tests search google book jackets

ConnorSheremeta:
- Added tests for google book jackets displaying/not displaying
#1635


spec/features/search_google_book_jackets_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    fill_in 'q', with: 'Savages; a film'
+    click_button 'search'
+
+    expect(page).to have_css("img[src*='https://books.google.com/books?vid=ISBN']")
+  end
+
+  scenario 'Search results without an isbn do not have a google book jacket' do
+    visit '/'
+
+    fill_in 'q', with: 'The college Shakespeare; 15 plays and the sonnets'
```
pgwillia: What do you think about asserting that there is only one item found in each of these tests?  I wondered if there were other results there that could challenge the negative assertion especially but I see that these title searches brings up just one result.

Issue #1391, opened 2018-11-15
[Discovery] Add optional email address field to send correction form


seanluyk:
**Describe the solution you'd like**
Bib services has requested that an optional field for email address be added to the send correction form so they can follow up with users as needed.

PR #1693, opened 2019-07-09
Add optional email address field to send correction form

ConnorSheremeta:
- Added optional email address to model, controller, form, and notify email
#1391

Before/After
1391SS


app/views/corrections/new.html.erb
```diff
@@ -20,16 +20,16 @@
                     required: true,
                     placeholder: 'Enter any details on the error, typo or suggestion here...',
                     class: 'form-control' %>
+
+    <%= f.label :email, 'Email (optional)' %>
+    <%= f.text_area :email,
```
pgwillia: Wondering why you used `text_area` for this field?  I would have used the [email_field](https://guides.rubyonrails.org/form_helpers.html#other-helpers-of-interest) like this:
```suggestion
    <%= f.email_field :email,
                    required: false,
                    placeholder: "Enter your email if you'd like to receive a follow up response as needed",
                    class: 'form-control' %>
```

app/views/corrections/new.html.erb
```diff
@@ -20,16 +20,19 @@
                     required: true,
                     placeholder: 'Enter any details on the error, typo or suggestion here...',
                     class: 'form-control' %>
+
+    <%= f.label :email, 'Email (optional)' %>
+    <%= f.email_field :email,
+                    required: false,
+                    placeholder: "Enter your email if you'd like to receive a follow up response as needed",
+                    class: 'form-control' %>
+                    rows: 1,
```
pgwillia:
```suggestion
```

app/views/corrections/new.html.erb
```diff
@@ -20,16 +20,19 @@
...
+                    required: false,
```
pgwillia:
```suggestion
```

app/views/corrections/new.html.erb
```diff
@@ -20,16 +20,19 @@
...
+                    placeholder: "Enter your email if you'd like to receive a follow up response as needed",
```
pgwillia:
```suggestion
```

app/views/corrections/new.html.erb
```diff
@@ -20,16 +20,19 @@
...
+                    class: 'form-control' %>
```
pgwillia:
```suggestion
```

ConnorSheremeta:
I couldn't see what you suggested the placeholder to be changed to so I changed it to what I think you had suggested

pgwillia:
> I couldn't see what you suggested the placeholder to be changed to so I changed it to what I think you had suggested

I thought the placeholder was fine as is.  It shadows the language used in the issue.

app/assets/stylesheets/ualib-custom.scss
```diff
@@ -2106,6 +2106,9 @@ form.range_limit input.range_begin, form.range_limit input.range_end {
   width: 100%;
 }
 
+form.new_correction {
```
pgwillia: I saw this in another of your pull requests.  Why do we need to make these forms 100% width?

ConnorSheremeta: Since the bootstrap upgrade the forms don't default a full width like they used to for some reason. They're still completely usable but just not as they used to be

Issue #1727, opened 2019-07-29
[Discovery] Add feature tests for sending a correction


ConnorSheremeta:
Add feature tests surrounding sending a correction for an item
From epic #1330

PR #1729, opened 2019-07-30
Add feature tests for sending a correction

ConnorSheremeta:
- Ensured that a user can send a correction
#1727


spec/features/send_correction_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      visit '/catalog/1002481'
+      within('.show-tools') do
+        click_link 'Send Correction'
+      end
+
+      fill_in 'correction[message]', with: 'correction message'
+      expect { click_button 'Submit Error, Typo or Suggestion' }
+        .to change { ActionMailer::Base.deliveries.count }.by(1)
+
+      expect(page).to have_content 'Thank you! We have received your correction and will be reviewing it soon.'
```
pgwillia: I probably haven't been paying enough attention to this so there's cleanup to do elsewhere, but we could DRY this by using the I18n value instead of copying the expected text.
```suggestion
      expect(page).to have_content I18n.t('request_form_success')
```

ConnorSheremeta:
The reason it was failing is because the text that is displayed after sending a correction isn't in i18n right now, let me know if you think the string should be added to i18n as part of this issue

Issue #1728, opened 2019-07-30
Add additional text for BPSC request form


weiweishi:
Update the label used on the request form to provide some additional information regarding the location's planned closures:

`First date of planned visit (open Mon-Fri 12-4:30pm, but renos are expected to cause intermittent closures in 2019-20 so please confirm hours here)`

PR #1730, opened 2019-07-31
Add additional text for BPSC request form

ConnorSheremeta:
- Added closure disclosure to form field
#1728

Screenshot from 2019-08-01 09-51-46

### DIFF
--- a/app/views/bpsc_read_on_site_requests/new.html.erb
+++ b/app/views/bpsc_read_on_site_requests/new.html.erb
@@ -8,7 +8,7 @@
   <%= bootstrap_form_for(@bpsc_read_on_site_request) do |f| %>
     <%= f.text_field :name, required: true %>
     <%= f.email_field :email, required: true %>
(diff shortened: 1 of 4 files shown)


app/views/bpsc_read_on_site_requests/new.html.erb
```diff
@@ -8,7 +8,7 @@
   <%= bootstrap_form_for(@bpsc_read_on_site_request) do |f| %>
     <%= f.text_field :name, required: true %>
     <%= f.email_field :email, required: true %>
-    <%= f.text_field :appointment_time, label: t('.date_of_planned_visit'), required: true %>
+    <%= f.text_field :appointment_time, label: t('.date_of_planned_visit').html_safe, required: true %>
```
pgwillia: `html_safe` is [generally not recommended](https://devdocs.io/rails~5.2/string#method-i-html_safe), but I'm good with this use because we control the content.

weiweishi: Good to know!

Issue #1367, opened 2018-11-01
Clipping of 264$c removes useful information


seanluyk:
**Describe the bug**
In cases where the publication information recorded in 264$c contains anything other than digits, information about questionable dates, date errors and BC dates are not displayed. This information is useful to researchers. One of the [index scripts](https://github.com/ualbertalib/discovery/tree/master/config/SolrMarc/index_scripts) is cleaning the dates in this field, which makes it possible to limit by date, etc. 

**To Reproduce**
See some of the following examples:

https://www.library.ualberta.ca/catalog/8343787 Should be: [2032 B.C. or 2031 B.C.?]
https://www.library.ualberta.ca/catalog/8343785 Should be: [2043 B.C.?]
https://www.library.ualberta.ca/catalog/1901116 Should be: 1957 [ie.1958]

**Expected behavior**
 - This information should be displayed somewhere
 - First date in the string should still be used for limiting/faceting  

**Additional context**
 - This is an issue with the source data, so may be difficult to deal with on the BL end
 - Part of a larger issue of non-dates being in date fields

ConnorSheremeta:
I see in the [third example above](https://www.library.ualberta.ca/catalog/1901116) that the publication date is in 260c (unlike the others where it is in 264c.) I just wanted to confirm with @redlibrarian that both fields should be displayed here, comma separated if both fields exist

weiweishi:
@redlibrarian I don't know if you've answered @ConnorSheremeta's question via another channel. Would you mind confirming if you'd like to display both 260c and 264c if they both exist? I don't know how likely this is to happen.

theLinkResolver:
Has anyone from cataloguing provided useful/actionable information on how these fields work? @weiweishi's question suggests there's a gap, but I am trying not to speak out of turn anymore..

weiweishi:
@theLinkResolver Thanks Scott. some explanation on how 260 and 264 are used in your practice would be helpful. I think in this particular case, we are wondering when 260c is used vs when 264c is used (some examples in Sean's request use 260c, while others use 264c). And we wonder if these fields/subfields can co-exist on the same record?

theLinkResolver:
@weiweishi Great! Happy to help. I will write up something soon.

theLinkResolver:
**Why we have 2:** 
- 260 is a legacy/old way of recording publication, distribution, etc. and does not distinguish other functions
[...]

PR #1736, opened 2019-08-07
Made display of publication year field not remove information

ConnorSheremeta:
- Stop using getDate index script for publication date display field
Fixes: #1367

Before:
Screenshot from 2019-08-07 14-26-21

After:
Screenshot from 2019-08-07 14-26-32


config/SolrMarc/index.properties
```diff
@@ -55,6 +55,7 @@ published_display = custom, removeTrailingPunct(260a)
 published_vern_display = custom, getLinkedField(260a)
 # used for facet and display, and copied for sort
 pub_date = custom, getDate
+pub_date_display = 260c:264c
```
pgwillia: The issue referenced 264$c.  Why both 260c and 264c?

app/controllers/catalog_controller.rb
```diff
@@ -172,7 +172,7 @@ def show
     config.add_show_field 'publisher_tesim', label: 'Publisher'
     config.add_show_field 'published_display', label: 'Published'
     config.add_show_field 'published_vern_display', label: 'Published'
-    config.add_show_field 'pub_date', label: 'Year'
+    config.add_show_field 'pub_date_display', label: 'Year'
```
pgwillia: Nice use of the display dynamic field!
https://github.com/ualbertalib/blacklight_solr_conf/blob/f0d3cd891c004e3e799d37d45c196e5970b35fb3/solr/blacklight-core/conf/managed-schema#L718

config/SolrMarc/index.properties
pgwillia: This doesn't appear to be working.  Nothing appearing in the `pub_date_display` field after indexing and the field isn't appearing
Screenshot from 2019-08-08 10-48-49

ConnorSheremeta: I'll confirm with Sam that this ticket involves both 260c and 264c, it appears to me that the publication date could be stored in either. I believe that your screenshotted example isn't working because utilising the SOLR_URL variable makes it impossible to re-index the data, only when rake ingest[] is run is when the data is re-indexed (this is my understanding.) The example I've screenshotted is titled "Church Missionary Society archive. Section III, Central records [microform]" which is part of the test data.

pgwillia: Unfortunately that doesn't explain it.  My solr_url is the local one 
```
pjenkins@DIG002TJ1:~/Code/discovery$ bundle exec rails c
Loading development environment (Rails 4.2.11.1)
[1] pry(main)> Rails.application.secrets.solr_url
=> "http://localhost:8983/solr/discovery"
```
And I purposefully ran the rake ingest tasks.  What I realize now is that the *_display dynamic type isn't indexed so my search for `pub_date_display:*` doesn't work.  The field is there if I search for an id that has that field like your 4072071.

Your example (catalog/4072071) does work as expected but the ones from the ticket don't: 
- catalog/8343787
- catalog/8343785
- catalog/1901116

pgwillia: These are all 264$c

pgwillia: Screenshot from 2019-08-08 13-58-28
Even the third one which is 260$c

ConnorSheremeta: Strange, the three from the ticket are working for me, not too sure what's going on... especially if 4072071 is working for you but the others aren't
Screenshot from 2019-08-09 08-49-29
Screenshot from 2019-08-09 08-49-10

pgwillia: Weird.  I'll look at what's going wrong on my machine.

pgwillia: > Weird. I'll look at what's going wrong on my machine.

Today there is nothing going wrong on my machine.  :100:

PR #1805, opened 2019-10-18
Update uat docker

pgwillia:
This will bring the dockerized discovery into line with how jupiter was done.
Uses
- modern ruby docker image
- puma instead of passenger

Works-around https://github.com/ualbertalib/discovery/issues/1796

To test on your machine
```
diff --git a/config/environments/uat.rb b/config/environments/uat.rb
index c9c78b9b..8c719409 100644
--- a/config/environments/uat.rb
+++ b/config/environments/uat.rb
@@ -1,4 +1,4 @@
-Rails.application.routes.default_url_options = { host: 'uat.library.ualberta.ca', port: 3002 }
+Rails.application.routes.default_url_options = { host: 'localhost', port: 3002 }
 
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
diff --git a/docker-compose.yml b/docker-compose.yml
index 195dd253..f460521c 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,8 @@ services:
   # Rails
   web:
     restart: always
-    image: ualbertalib/discovery:uat
+    build:
+      context: .
     volumes:
       - assets:/app/public/assets/
     command: bundle exec puma -e uat
```
Then you can
```
$ docker-compose build 
$ docker-compose up -d
```

I did notice an oddity which I'll continue to investigate
Screenshot from 2019-10-18 11-57-51


pgwillia:
This is what I would have expected it to look like
Screenshot from 2019-10-18 12-08-03

pgwillia:
This hint from https://github.com/ualbertalib/di_internal/blob/master/System-Adminstration/UAT-Environment.md#deploy-jupiter
> Have assest changes? Are they not showing up? deploy.sh will run rake assests:precompile for you but rails may still be caching the references to the old assest versions. To clear the cache, restart the rails server. This can be done with the following:

```
docker-compose exec web bash
touch tmp/restart.txt
```

worked to clear up the oddity.

PR #1807, opened 2019-10-22
Add version file and meta generator tag

### LINKED ISSUE #1343: Add version.rb file for Discovery
**Is your feature request related to a problem? Please describe.**
It's hard to know which version of code you're looking at.

**Describe the solution you'd like**
- [x] Can use example in PMPY for the version.rb file as a constant
- [x] Embedded into the HTML meta-tag
- [ ] Manage the version through a rake task, to update the version.rb

**Additional context**
Similar to https://github.com/ualbertalib/jupiter/issues/55

pgwillia:
It's hard to know which version of the code you're looking at.
If we can keep this up-to-date in our release management then we'll be
able to tell which version we see deployed by checking the page source.

Screenshot from 2019-10-22 10-33-47

fixes #1343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 
 ## [Unreleased]
 
+### Added
+- Add version file and meta generator tag [#1343](https://github.com/ualbertalib/discovery/issues/1343)
+
 ### Changed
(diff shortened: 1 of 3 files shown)


pgwillia:
> Looks good. I'm curious what measures can be done to ensure that this is changed every release. I noticed that updating Jupiter's version sometimes gets forgotten about

Me too.  Maybe a release checklist?  Or a release pr template?  I don't think the auto generation we built into jupiter works well enough which is why I excluded that part here.

PR #1801, opened 2019-10-15
add a migration which will change db/table/col to utf8mb encoding

pgwillia:
Changes the database defaults as well as any existing tables and columns.

Before:
```
mysql> SELECT @@character_set_database, @@collation_database;
+--------------------------+----------------------+
| @@character_set_database | @@collation_database |
+--------------------------+----------------------+
| latin1                   | latin1_swedish_ci    |
+--------------------------+----------------------+
1 row in set (0.00 sec)

mysql> show create table searches;
+----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table    | Create Table                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| searches | CREATE TABLE `searches` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `query_params` text COLLATE utf8_unicode_ci,
  `user_id` int(11) DEFAULT NULL,
  `user_type` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `created_at` datetime DEFAULT NULL,
  `updated_at` datetime DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `index_searches_on_user_id` (`user_id`)
) ENGINE=InnoDB AUTO_INCREMENT=32 DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci |
+----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```
The migration:
```
pjenkins@DIG002TJ1:~/Code/discovery$ bundle exec rake db:migrate
== 20190314159265 ChangeToUtf8mbEncoding: migrating ===========================
-- execute("ALTER DATABASE `discovery_development` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;")
   -> 0.0552s
-- execute("ALTER TABLE `bookmarks` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;")
   -> 0.1103s
[...]

PR #1810, opened 2019-10-24
changed migration which will change db/table/col to utf8 encoding

pgwillia:
Was originally going to the best practice standard of utf8mb4 but scaled
back my ambition when we encounted column size too large. While we do want
CJK characters (maybe even emojis) this won't break anything that isn't
already broken.

Merge of this PR should result in immediate new release to replace 3.0.111.

#1800 similar to #1801


pgwillia:
> LGTM. Something I missed in your other PR... what's the significance of 'rolling' over the version number? i.e. listing .12 rather than .112

Thanks for your keen eyes.  .12 vs .112 was a mistake.  I've corrected it.

PR #1901, opened 2020-02-07
Tests refactor

pgwillia:
I was running the specs over and over and getting annoyed by the deprecations.  I was able to address some easily.  There are others related to Blacklight 6.0 deprecations that will need to be addressing in Blacklight. 

- I remove deprecated chromedriver-helper in favour of webdriver and upgraded capybara for `Selenium::WebDriver::Error::InvalidElementStateError is deprecated. Use Selenium::WebDriver::Error::ElementNotInteractableError (ensure the driver supports W3C WebDriver specification) instead.`
- Updated instances of 'should' to the expect matcher. 
```
Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` with `config.expect_with(:rspec) { |c| c.syntax = :should }` instead.
```
- Used #[] instead of get `get is deprecated and will be removed from a future release (Use #[] or #fetch instead).`
- Removed redundant params from call to kaminari `Explicitly passing params to helpers could be omitted.`


app/views/catalog/email.html.erb
```diff
@@ -1,43 +1,45 @@
...
-<% end %>
+            <% @documents.each do |doc| %>
+                <%= hidden_field_tag 'id[]', doc[:id] %>
+                    <% end %>
+                        <%- if params[:sort] -%>
+                            <%= hidden_field_tag 'sort', params[:sort] %>
+                                <%- end -%>
+                                    <%- if params[:per_page] -%>
+                                        <%= hidden_field_tag 'per_page', params[:per_page] %>
+                                            <%- end -%>
```
pgwillia: Rats!  Using html beautifier didn't work as well as I had hoped for erb.  I'll fix this.

PR #1914, opened 2020-02-13
Revert "Bump vcr from 5.0.0 to 5.1.0"

pgwillia:
Reverts ualbertalib/discovery#1903

Discussion on #general-dev slack lead to questions of if we can accept the Hippocratic License in our projects.  Reverting and pinning until we can properly address this question.


Gemfile.lock
```diff
@@ -395,7 +395,7 @@ GEM
       unf_ext
     unf_ext (0.0.7.6)
     unicode-display_width (1.6.1)
-    vcr (5.1.0)
+    vcr (5.0.0)
```
lagoan: Would we also want to change the Gemfile file ?

Issue #1843, opened 2019-11-08
Set up data synchronization between Symphony and Discovery DB


weiweishi:
Since release 3.0.113, we have moved the management of location codes, circulation rules, statuses, and item types into DB, instead of static yml file in Discovery. This will avoid a new release every time a change is introduced in Symphony, and will allow for more efficient/accurate data used in Discovery.

Now we need to work with ILS team to create a solution that will sync the data from Symphony to  the DB, possibly as a cron job:
- [x] Meeting among ILS/Dev/@redlibrarian to talk about possible ways to get the information out of Symphony
- [x] Whiteboarding a solution that will work efficiently, with less development/system administration effort involved. 
- [x] Develop/Implement the solution.

pgwillia:
More information https://github.com/ualbertalib/discovery/issues/1760

weiweishi:
Reports from ILS team is ready to go. We will need a sample file to create script, establish workflow with @nmacgreg. @pgwillia will follow up with ILS for the reports.

pgwillia:
We've tested data synchronization between Symphony and Discovery DB.  Happy with results!

PR #1936, opened 2020-03-12
Update README with symphony_nightly task

pgwillia:
#1843

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Integration tests (run against <http://search-test.library.ualberta.ca/>)
 ## Additional Tasks for uat, staging and production
 
 1.  `bundle exec rake assets:precompile` this can take several minutes
-2.  create cron jobs to ingest `bundle exec rake ingest[sfx]`, `bundle exec rake ingest[databases]` and clean session table `bundle exec rake sessions:cleanup`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,3 +61,5 @@ export SOLR_INGEST_URL=http://localhost:8983/solr/your-new-solr-collection
 bundle exec rake ingest[collection]
 unset SOLR_INGEST_URL # if desired
 ```
+
+To populate the human readable strings for the 'Where is this?' table.  Assuming `./data/data4discovery.txt` and `./data/Data4DiscoveryManual.txt` are present then you can invoke `bundle exec rails g symphony_nightly  && bundle exec rake db:migrate` to populate the tables with the latest configurations. If for some reason this doesn't go well, `bundle exec rake db:rollback` and delete the migration at `db/migrate/<todays date>_symphony_nightly_<todays date>.rb` will revert to a known good state.
```
murny: `'Where is this?'` haha is this to be filled in later?

pgwillia: 😆 No, that's what the title of the table is. 
image
 It's also known as the holdings table.  I wasn't sure what to call it.  Suggestions?

murny: Ahhh that makes sense 😅. Thanks for explanation 👍

PR #1937, opened 2020-03-12
use memoization to protect against N+1 query

pgwillia:
Was doing a database lookup for each holding of an item.  Where the item
has volumes upon volumes like catalog/1772030 this is very inefficient.

The solution incorporates memoization using `Hash.new` initializer to call a block only if the entry doesn't already
exist (https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/#and-what-about-parameters?)

I was worried that instance variables (@Locations) would have a cache invalidation
problem, but found out that there is a seperate controller instance for
each web request so this isn't really an issue.

### DIFF
--- a/app/helpers/holdings_helper.rb
+++ b/app/helpers/holdings_helper.rb
@@ -17,27 +17,36 @@ module HoldingsHelper
   end
 
   def symphony_status(item)
-    Status.find_by!(short_code: item[:status].downcase.underscore).name
-  rescue ActiveRecord::RecordNotFound => e
-    Rollbar.error("Error retriving name for Status #{item[:status].downcase.underscore}", e)
-    Status.create(short_code: item[:status].downcase.underscore, name: 'Unknown')
-    'Unknown'
+    @symphony_status ||= Hash.new do |h, key|
+      status = Status.find_by(short_code: key) || begin
+        Rollbar.error("Error retriving name for Status #{key}", e)
+        Status.create(short_code: key, name: 'Unknown')
+      end
+      h[key] = status.name
+    end
+    @symphony_status[item[:status]]
   end
 
   def item_type(item)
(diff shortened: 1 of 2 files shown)


app/helpers/holdings_helper.rb
```diff
@@ -17,27 +17,36 @@ def holdings(document, method)
...
-    Status.create(short_code: item[:status].downcase.underscore, name: 'Unknown')
-    'Unknown'
+    @symphony_status ||= Hash.new do |h, key|
+      status = Status.find_by(short_code: key) || begin
+        Rollbar.error("Error retriving name for Status #{key}", e)
+        Status.create(short_code: key, name: 'Unknown')
+      end
+      h[key] = status.name
+    end
+    @symphony_status[item[:status]]
```
pgwillia: The short codes aren't modified on entry so they won't need to be modified on lookup anymore either.
https://github.com/ualbertalib/discovery/blob/8dc22c17b407b42d551b3714ec39e6c16725db1c/lib/generators/symphony_nightly/templates/update_symphony_nightly.erb#L32-L35

app/helpers/holdings_helper.rb
```diff
@@ -17,27 +17,36 @@ def holdings(document, method)
...
-    Rollbar.error("Error retriving name for Status #{item[:status].downcase.underscore}", e)
-    Status.create(short_code: item[:status].downcase.underscore, name: 'Unknown')
-    'Unknown'
+    @symphony_status ||= Hash.new do |h, key|
+      status = Status.find_by(short_code: key) || begin
+        Rollbar.error("Error retriving name for Status #{key}", e)
+        Status.create(short_code: key, name: 'Unknown')
+      end
+      h[key] = status.name
+    end
```
mbarnett: This is honestly, very much cleverer than my suggestion 👍. Learned a new trick today.

Issue #1939, opened 2020-03-13
Refactor for SolrMarc


pgwillia:
Since release 3.0.113, we have moved the management of location codes, circulation rules, statuses, and item types into DB, instead of static yml file in Discovery. This will avoid a new release every time a change is introduced in Symphony, and will allow for more efficient/accurate data used in Discovery.

Now we need to use this same information when indexing/ingesting the MARC data into Solr. As part of the [symphony_nightly generator](https://github.com/ualbertalib/discovery/blob/master/lib/generators/symphony_nightly/symphony_nightly_generator.rb) we could update these files
- [x] [location_facet.bsh](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/index_scripts/location_facet.bsh)
- [x] [institution.bsh](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/index_scripts/institution.bsh)

Should check if this should get the same treatment
- ~[format.bsh](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/index_scripts/format.bsh)~
- ~[electronic_map.properties](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/translation_maps/electronic_map.properties)~

#1760

pgwillia:
This is in production waiting for confirmation that it's working in the wild.

pgwillia:
Been through nightly and monthly cycles.  Closing

PR #1944, opened 2020-03-26
add solr_marc_prep generator consumed by the ingest rake task

pgwillia:
Since release 3.0.113, we have moved the management of location codes,
circulation rules, statuses, and item types into DB, instead of static
yml file in Discovery. This will avoid a new release every time a change
is introduced in Symphony, and will allow for more efficient/accurate
data used in Discovery.

Now we need to use this same information when indexing/ingesting the
MARC data into Solr. When `rake ingest[collection]` is invoked, the
generator will be called which will use the current database values in
a template to replace two files that map to human readable strings for
Institution and Library in the application's facets.

#1939


pgwillia:
https://github.com/ualbertalib/discovery/pull/1966 replaces this.

PR #1945, opened 2020-03-27
Add a simple standard PR template

pgwillia:
## Context
We've been using this in Oaisys for a bit, and we've found it handy elsewhere.  It was just [added to Jupiter](https://github.com/ualbertalib/jupiter/pull/1570)

The intention isn't that this should be super onerous or require a lot of extra "paperwork" from people, it's just intended to remind people to be kind to reviewers by giving them a bit of context to help them understand your PR.

Things are always optional where they don't really apply – like there's no issue number associated with this PR, and it would be pointless busywork to go create one when I just had the idea to add this five minutes ago.

## What's New
- Adds a basic template GitHub will use to pre-populate PR descriptions.
- That's it. That's the PR.


pgwillia:
> Oh wait, I think you have it in the wrong directory? I think it needs to be directly under .github?

Doh!  I tried to do it all in the github UI rapidly and failed. Let me try again.

PR #1950, opened 2020-03-31
remove danger

pgwillia:
Found that travis was exposing danger token.  We changed the token and the CI for jupiter.  Just going to remove danger here.


pgwillia:
I'll update the token.

PR #1966, opened 2020-04-22
add solr_marc_prep generator consumed by the ingest rake task

pgwillia:
Since release 3.0.113, we have moved the management of location codes,
circulation rules, statuses, and item types into DB, instead of static
yml file in Discovery. This will avoid a new release every time a change
is introduced in Symphony, and will allow for more efficient/accurate
data used in Discovery.

Now we need to use this same information when indexing/ingesting the
MARC data into Solr. When `rake ingest[collection]` is invoked, the
generator will be called which will use the current database values in
a template to replace two files that map to human readable strings for
Institution and Library in the application's facets.

#1939

### DIFF
--- a/lib/generators/solr_marc_prep/templates/institution.bsh.erb
+++ b/lib/generators/solr_marc_prep/templates/institution.bsh.erb
@@ -10,14 +10,13 @@ public Set getInstitutions(Record record){
   if (code_string != null){
   String[] codes = code_string.split(" ");
   Hashtable institution_codes = new Hashtable();
-
-  <%- Location.order(:symphony_id).each do |l| %>
-  <%-   if l.symphony_id.present? && l.library -%>
+  <% Location.includes(:library).order(:symphony_id).where.not(symphony_id: nil).each do |l| -%>
+  <%-   if l.library %>
   <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.library.name}\");" -%>
-  <%-   elsif l.symphony_id.present? -%>
+  <%-   else %>
   <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.name}\");" -%>
   <%-   end -%>
-  <%- end -%>
+  <%- end %>
 
  for (String code : codes){
     institutions.add(institution_codes.get(code));
--- a/lib/generators/solr_marc_prep/templates/institution.bsh.erb
+++ b/lib/generators/solr_marc_prep/templates/institution.bsh.erb
@@ -1,0 +1,27 @@
+import org.marc4j.marc.Record;
+
+org.solrmarc.index.SolrIndexer indexer = null;
+
+public Set getInstitutions(Record record){
+
+  Set institutions = new LinkedHashSet();
+  String code_string = indexer.getFirstFieldVal(record, "596a");
+
+  if (code_string != null){
+  String[] codes = code_string.split(" ");
+  Hashtable institution_codes = new Hashtable();
+
+  <%- Location.order(:symphony_id).each do |l| %>
+  <%-   if l.symphony_id.present? && l.library -%>
+  <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.library.name}\");" -%>
+  <%-   elsif l.symphony_id.present? -%>
+  <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.name}\");" -%>
+  <%-   end -%>
+  <%- end -%>
+
+ for (String code : codes){
+    institutions.add(institution_codes.get(code));
+  }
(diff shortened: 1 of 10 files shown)


lib/tasks/ingest.rake
```diff
@@ -26,8 +26,13 @@ task :ingest_info do
   puts "Solr collection contains #{response['response']['numFound']} results."
 end
 
+desc 'update maps for location facets from current db state'
+task :update_maps do
+  sh 'rails g solr_marc_prep'
```
pgwillia, line 31: This will update with the latest values in the database before ingest is started.

spec/lib/generators/solr_marc_prep/solr_marc_prep_generator_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  it "creates a institutions beanshell mapping script" do
+    Location.create(short_code: 'UAINTERNET', symphony_id: '44', name: "University of Alberta Internet")
+    run_generator
+    assert_file "config/SolrMarc/index_scripts/institution.bsh", "institution_codes.put("44", "University of Alberta");"
+  end
+
+  it "creates a location facets beanshell mapping script" do
+    assert_file "config/SolrMarc/index_scripts/location_facet.bsh"
```
pgwillia, line 19: Ought to check that the file changes but couldn't find a way to do that.

lib/generators/solr_marc_prep/templates/location_facet.bsh.erb
```diff
@@ -0,0 +1,18 @@
...
+org.solrmarc.index.SolrIndexer indexer = null;
+
+public Set getLocationFacet(Record record){
+
+  Set locations = new LinkedHashSet();
+  Set codes = indexer.getFieldList(record,"926m");
+  Hashtable location_codes = new Hashtable();
+  <% Location.order(:short_code).each do |l| %>
+    <%= "location_codes.put(\"#{l.short_code}\", \"#{l.name}\");" -%>
+  <% end %>
```
lagoan, line 12: If you write your template in https://github.com/ualbertalib/discovery/pull/1966/files#r413412089 as you did here, would you get the same leading spaces ?

lib/generators/solr_marc_prep/templates/institution.bsh.erb
```diff
@@ -0,0 +1,27 @@
...
+  String[] codes = code_string.split(" ");
+  Hashtable institution_codes = new Hashtable();
+
+  <%- Location.order(:symphony_id).each do |l| %>
+  <%-   if l.symphony_id.present? && l.library -%>
+  <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.library.name}\");" -%>
+  <%-   elsif l.symphony_id.present? -%>
+  <%=     "institution_codes.put(\"#{l.symphony_id}\", \"#{l.name}\");" -%>
+  <%-   end -%>
+  <%- end -%>
```
lagoan, line 20: In both of the if statements we check if symphony_id is present, would Locations like the Kule collections be needed here ? If not could we just filter out the Locations where symphony is present on the query ?

lib/generators/symphony_nightly/templates/update_symphony_nightly.erb
```diff
@@ -27,12 +27,13 @@ class UpdateSymphonyNightly<%= Time.now.strftime('%Y%m%d') %> < ActiveRecord::Mi
     create_table_libraries
     create_table_locations
 
+    Location.create(short_code: 'bmufa', name: "Kule Collection")
```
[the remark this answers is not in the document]
lagoan: You are right in placing it here in order to make sure it exist after every update.

lib/tasks/ingest.rake
```diff
@@ -26,8 +26,13 @@ task :ingest_info do
   puts "Solr collection contains #{response['response']['numFound']} results."
 end
 
+desc 'update maps for location facets from current db state'
+task :update_maps do
```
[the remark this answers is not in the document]
lagoan: Naming things is hard! How about ``update_solr_marc_maps`` ?

lib/generators/solr_marc_prep/templates/location_facet.bsh.erb
pgwillia: A little more playing seemed to clear this up.  Thanks for pushing me to do it.

lib/generators/solr_marc_prep/templates/institution.bsh.erb
pgwillia: Good thinking!  I did the refactor and I think it improves the readability.

PR #1994, opened 2020-05-14
Add Hathitrust ETAS access

ConnorSheremeta:
#1973
- Added controller to handle the modal
- Migration to create Hathitrust table
- Rake tasks to populate and clear the table
Screenshot from 2020-05-14 13-48-13
Screenshot from 2020-05-14 13-48-07


app/assets/javascripts/ualib-discovery.js
```diff
@@ -39,3 +39,38 @@ window.onload=setiFrameAttr;
...
+      jsonpCallback: "hathitrustResponse"
+    });
+  });
+}
+
+function hathitrustResponse(json) {
+  if (!jQuery.isEmptyObject(json) && !jQuery.isEmptyObject(json.records)) {
+
+    // Proper response exists, show button and set hidden div text.
+    $('#hathitrustButton').css('display', 'block');
```
pgwillia, line 58: This is the magic.

app/assets/javascripts/ualib-discovery.js
```diff
@@ -39,3 +39,38 @@ window.onload=setiFrameAttr;
...
+    $hathitrust_link_div.text(recordURL + '?urlappend=%3Bsignon=swle:wayf');
+  }
+}
+
+function hathitrustModal() {
+  // Retrieve link and set form action.
+  var $hathitrust_link_div = $('#hathitrustLink');
+
+  var $hathitrust_form = $('#hathitrustForm');
+  $hathitrust_form.attr('action', $hathitrust_link_div.text());
```
pgwillia, line 71: The form doesn't have a [form method](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method) (GET, POST) so the default [GET](https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data#The_GET_method) is used.  You'll need to add a [hidden input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden) for the `urlappend` part.

As a result I don't think the parameters, the `urlappend`... part sticks.  So the user doesn't get the redirect to select their idp.

app/assets/javascripts/ualib-discovery.js
```diff
@@ -39,3 +39,38 @@ window.onload=setiFrameAttr;
...
+
+function hathitrustResponse(json) {
+  if (!jQuery.isEmptyObject(json) && !jQuery.isEmptyObject(json.records)) {
+
+    // Proper response exists, show button and set hidden div text.
+    $('#hathitrustButton').css('display', 'block');
+    var recordURL = Object.values(json.records)[0].recordURL;
+
+    var $hathitrust_link_div = $('#hathitrustLink');
+    $hathitrust_link_div.text(recordURL + '?urlappend=%3Bsignon=swle:wayf');
```
pgwillia, line 62: There was a bit more discussion on the ticket on how to take the user straight to the UofA authentication. Not sure if that's the plan at this stage.

lib/tasks/hathitrust.rake
```diff
@@ -0,0 +1,15 @@
+# rubocop:disable Style/HashSyntax
+desc 'Populate the hathitrust table from a csv'
+task :populate_hathitrust_table => :environment do
```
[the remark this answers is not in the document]
mbarnett: We should definitely look at performance improvements, but it's loading somewhere north of 1.1 million rows into the DB, so it's not going to be the fastest thing on earth. It _should_ be fine to run this with the web app up, with the impact being that HathiTrust access slowly spreads to more and more items.

app/assets/javascripts/ualib-discovery.js
mbarnett: I had meant to throw this into my review, but I don't think the modal should be a form, it should be fine if it's just a div or whatever – "Accept" can just be an `<a>` tag pointing at the recordURL and styled to look like a button, "Cancel" can just close the modal.

app/assets/javascripts/ualib-discovery.js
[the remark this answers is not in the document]
pgwillia: I don't see an itemURL in the response:
```
json.records: {…}
  008466190: {…}
    isbns: (8) […]​
    issns: []
    lccns: (1) […] 
    oclcs: (1) […]
    publishDates: (1) […]
    recordURL: "https://catalog.hathitrust.org/Record/008466190"
​​    titles: (1) […]
```

app/models/hathitrust.rb
```diff
@@ -0,0 +1,2 @@
+class Hathitrust < ActiveRecord::Base
```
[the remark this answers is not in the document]
pgwillia: I had trouble with this too.  I like @mbarnett suggestion `HathiTrustOverlapRecord`

app/assets/javascripts/ualib-discovery.js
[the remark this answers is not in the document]
mbarnett: So the full response for that record is:

```{"records":{"008466190":{"recordURL":"https:\/\/catalog.hathitrust.org\/Record\/008466190","titles":["American nursing : a history of knowledge, authority, and the meaning of work"],"isbns":["9780801895654","9780801895654","9780801895647","9780801895647","9780801895654","9780801895654 (pbk : alk paper)","9780801895647","9780801895647 (hardcover : alk paper)"],"issns":[],"oclcs":["436259144"],"lccns":["2009037380"],"publishDates":["2010"]}},"items":[{"orig":"University of California","fromRecord":"008466190","htid":"uc1.31822037490497","itemURL":"https:\/\/hdl.handle.net\/2027\/uc1.31822037490497","rightsCode":"ic","lastUpdate":"20191124","enumcron":false,"usRightsString":"Limited (search-only)"}]}```

The "record" is the metadata, which has an array, items, which are the physical or electronic resource you can check out. Inside the item array entry there's an itemURL (https://hdl.handle.net/2027/uc1.31822037490497) that takes you to the actual "sign out this item via the Emergency HathiTrust Access Service" page.

I think that itemURL is what we want to be linking to, because that's what we're offering them with the button, and if we just send them to the HathiTrust catalog record, it's not clear how to get from there to full text access.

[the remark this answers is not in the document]
Comment (mbarnett) [reply app/assets/javascripts/ualib-discovery.js:62]: Here's how HathiTrust explains the fields in the "Important fields from API results" bit of https://www.hathitrust.org/how-to-add-etas-records-to-your-catalog

```
Catalog Record URL: ‘recordURL’
The URL for the catalog page.

Item Record URL: ‘itemURL’
The URL for a specific digital item
```

I think we want to provide people to the item, not the catalog page.

File: db/migrate/20200508004437_create_hathitrusts.rb [hunk 6]
```diff
@@ -0,0 +1,10 @@
+class CreateHathitrusts < ActiveRecord::Migration
+  def change
+    create_table :hathitrusts do |t|
+      t.integer :oclc, null: false
+      t.integer :local_id, null: false, primary_key: true, index: true
+
+      t.timestamps null: false
```
mbarnett, line 7: I'd say we could get away without the timestamps, to save some space on the table. Might even be worth using local_id as the table id and saving another few bytes that way. With 1.whatever million rows, it adds up.

lib/tasks/hathitrust.rake
```diff
@@ -0,0 +1,15 @@
...
+  require 'csv'
+  CSV.open(Rails.root.join('lib', 'tasks', 'files', '20200508_hathitrust_overlap_report.csv'), 'r',
+           col_sep: "\t", headers: true).each do |row|
+    Hathitrust.create!(row.to_h)
```
mbarnett, line 7: We'll want to account for the possibility of impatience or other issue causing the task to be aborted halfway through – we don't want to add a ton of duplicate rows. `find_or_create` would be better here.

I'd say (especially if you turn the table into just the two columns, id and oclc) it would even be worth experimenting with hand-crafting the sql insert (or more properly, upsert https://www.postgresqltutorial.com/postgresql-upsert/) statements, to see if that results in better performance by bypassing creating a million plus ruby objects that the task doesn't actually use (meaning we're just creating pressure on the garbage collector as it periodically has to sweep up all of these objects that get created).

lib/tasks/hathitrust.rake
```diff
@@ -0,0 +1,15 @@
+# rubocop:disable Style/HashSyntax
+desc 'Populate the hathitrust table from a csv'
+task :populate_hathitrust_table => :environment do
+  require 'csv'
+  CSV.open(Rails.root.join('lib', 'tasks', 'files', '20200508_hathitrust_overlap_report.csv'), 'r',
+           col_sep: "\t", headers: true).each do |row|
```
mbarnett, line 6: We'll probably want to provide some feedback to the operator that the task isn't stuck or something (people have been jumpy about long running rake tasks in the past if it didn't provide some kind of indication it was still running). I'm thinking something like adding a counter you increment every row, `print '.'` every 100th row, and maybe a newline and updated total every 5000 rows, and a `finished!` when it's completed successfully.

This would lead to output like:

```
..................................................(5000)
..................................................(10000)
..................................................(15000)
etc
finished!
```

etc for a total of 200ish lines of output.

lib/tasks/hathitrust.rake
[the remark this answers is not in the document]
pgwillia: For information only
```
pjenkins@DIG002TJ5:~/Code/discovery$ time rake populate_hathitrust_table
	
real	286m8.806s
user	80m33.230s
sys	8m2.890s
```

app/assets/javascripts/ualib-discovery.js
```diff
@@ -39,3 +39,38 @@ window.onload=setiFrameAttr;
     Blacklight.doResizeFacetLabelsAndCounts();
   });
 })(jQuery);
+
+function getHathitrustLink(oclId) {
+  $(document).ready(function () {
+    $.ajax({
+      url: 'http://catalog.hathitrust.org/api/volumes/brief/oclc/' + oclId + '.json&callback=hathitrustResponse',
```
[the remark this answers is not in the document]
ConnorSheremeta: jQuery handles it for me! Removed.

app/views/hathitrust/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= javascript_tag "hathitrustModal()" %>
+
+<div class="modal-header col-12">
+  <h1 class="modal-title"><%= t('hathitrust.form.title') %></h1>
+  <button type="button" class="ajax-modal-close close" data-dismiss="modal" aria-hidden="true">×</button>
+</div>
+
+<form id="hathitrustForm" target="_blank" rel="noopener noreferrer">
+  <div class="modal-body">
+    <%= render partial: '/flash_msg' %>
```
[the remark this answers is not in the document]
ConnorSheremeta: Oops, leftover code from the email modal, removed!

lib/tasks/hathitrust.rake
```diff
@@ -1,15 +1,22 @@
...
+  desc 'Populate the hathitrust table from a csv'
+  task populate_table: :environment do
+    require 'csv'
+    CSV.foreach(Rails.root.join('lib', 'tasks', 'files', '20200508_hathitrust_overlap_report.csv'),
+                col_sep: "\t", headers: true).with_index do |row, index|
+
+      print '.' if ((index + 1) % 100).zero?
+      print "(#{index + 1})\n" if ((index + 1) % 5000).zero?
+
+      ActiveRecord::Base.connection.execute("INSERT IGNORE INTO hathitrust_overlap_records (catalog_id, oclc) VALUES
```
ConnorSheremeta, line 11: FYI this sped things up by ~30%!

mbarnett: Nice! It looks like we could probably do even better by using `LOAD DATA INFILE`: https://medium.com/@benmorel/high-speed-inserts-with-mysql-9d3dcd76f723, but it's not clear to me whether or not this would be restricted in prod. @pgwillia have you ever used that w/ Discovery at all?

mbarnett: Here's what I'm thinking: @ConnorSheremeta, you could add a second ingest task (`hathitrust:fast_populate_table` or something) that tries to do the `LOAD DATA INFILE` trick, and then have the task output whether or not it was successful by having it do a `select count(*)` on the table afterwards and verifying it has however many rows it should.

Then on the deploy we're basically saying "run the fast rake task and if it works, don't do anything further; otherwise run the regular ingest task but be aware it may take several hours to complete".

pgwillia: I have not used that with Discovery at all.  I suspect that it would be a no-go.  Would have to work with Neil to put the file at `/var/lib/mysql-files/` or to [enable ](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_local_infile) `local_infile`. I could ask Neil if this is hypothetically possible?

Like you mentioned earlier, now that the (now shorter) long running task has feedback the links to HathiTrust can trickle in gradually and the user won't be the wiser.

mbarnett: Ah, good point. Let's just go with this as is, and let the links trickle in. @ConnorSheremeta , don't worry about the load stuff.

lib/tasks/hathitrust.rake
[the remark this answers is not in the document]
pgwillia: The revised task
```
pjenkins@DIG002TJ5:~/Code/discovery$ time rake hathitrust:populate_table
..................................................(5000)
...
..................................................(1355000)
.............................................
Finished!

real	165m6.559s
user	14m7.820s
sys	2m53.567s
```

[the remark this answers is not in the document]
ConnorSheremeta: Wow that's quite faster! Thank you for doing this. I wish I would've ran benchmarks on every performance improvement to see how effective each was!

PR #2012, opened 2020-06-05
bump bundler to 2.1.4 and add rvm 2.6 to travis

pgwillia:
Bumping bundler because we've changed or deployment strategy and a modern bundler is used.

Adding 2.6 to travis to prepare us for the next Ruby update.


spec/spec_helper.rb
```diff
@@ -1,6 +1,5 @@
 require 'simplecov'
 SimpleCov.start 'rails'
-require_relative './blacklight_helper.rb'
```
pgwillia, line 3: I don't think that the `blacklight_helper.rb` is actually **doing** anything.

spec/support/test_thread_monkeypatch.rb
```diff
@@ -0,0 +1,16 @@
+# Rails 4.2 call `initialize` inside `recycle!`. However Ruby 2.6 doesn't allow calling `initialize` twice.
+# See for detail: https://github.com/rails/rails/issues/34790
```
pgwillia, line 2: Copied the monkey patch from the suggestions in the issue.

PR #2019, opened 2020-06-23
Header footer reskin

nnunn:
## Context

Fix to styling on advanced search page and Hathitrust buttons.

Related to # (issue)

## What's New

minor style fixes. Removed sticky footer css as we are no longer using it.


pgwillia:
The advanced search looks good:
image

So does the hathi trust link
image

PR #2021, opened 2020-06-24
fix regression test after reskin

pgwillia:
The reskin simplified the layout and the path to the list of libraries on the advanced search changed.

This makes the test a little bit more robust by matching relative to the form  element


mbarnett:
That makes sense. I was wondering who else knew perl.

pgwillia:
image

PR #2027, opened 2020-07-13
discourage holds placed on items with HathiTrust access

### LINKED ISSUE #2003: Placeholder issue for suppressing the Place Hold button when HathiTrust link is present.
This may not become necessary, as SLT still needs to make a decision on "curbside pickup", but that's likely to be decided on soon. I just want to capture the scope of this fix and take an initial stab at an easy way it would work.

Our initial understanding of HathiTrust's ETAS programme was that when we restored access to the print collection we would have to remove the HathiTrust ETAS links (i.e. ETAS is only allowed when there is no access to the print collection). Clarification from HathiTrust sent out on Friday indicates that access to the print collection can be restored as long as ETAS items are not accessible. Because curbside pickup will be staff mediated, we can ensure this happens by suppressing the Place Hold button on records that have a HathiTrust link.

I think the *simplest* solution is probably just as follows: if the [**hathitrustlink**](https://github.com/ualbertalib/discovery/blob/c0a3433bf32300b376b2a2ad692b821b7da2946d/app/views/catalog/_hathitrust_link.html.erb#L3) div is present, then suppress the [**btn hold** CSS class](https://github.com/ualbertalib/discovery/blob/c657e24288022849204f82d25de8d0f8dfbee907/app/views/catalog/_place_hold.html.erb#L1). There's another link in [_place_hold.html.erb](https://github.com/ualbertalib/discovery/blob/c657e24288022849204f82d25de8d0f8dfbee907/app/views/catalog/_place_hold.html.erb), but it doesn't have an identifier, so I'm not sure how best to suppress that link. This is just a suggestion, there may be better ways to implement this feature.

@mbarnett raised the question of users possibly figuring out how to construct a place hold link on their own to bypass the record link and place a hold that way. I don't think it's worth trying to build in the extra logic to prevent that at this stage.

pgwillia:
## Context
Our initial understanding of HathiTrust's ETAS programme was that when we restored access to the print collection we would have to remove the HathiTrust ETAS links (i.e. ETAS is only allowed when there is no access to the print collection). Clarification from HathiTrust indicates that access to the print collection can be restored as long as ETAS items are not accessible.

We're going to suppress the hold button in Discovery whenever the record is in the overlap table, without regard for whether or not the item is actually still available via the HathiTrust API, and knowing that the overlap report is not currently being refreshed in a timely manner.

To be clear: this means that we 100% expect that there will be corner cases in which the hold button will be unavailable but no access in HathiTrust button will be present, because the item is unavailable through their API. This is not a bug, this is a compromise we are making to do this in a timely manner – users are expected to contact public services to resolve access to those items. Weiwei will communicate this expectation to public services.

Resolves #2003 

## What's New

If the item is in our Hathi Trust overlap report suppress the "Hold" button.

with_hold_button
without_hold_button

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -61,7 +61,7 @@ class CatalogController < ApplicationController
     @document.delete('publisher_tesim') if @document['published_display']
 
     @document['oclc'] = HathitrustOverlapRecord.find_by(catalog_id: @document['id'])&.oclc
-  
+
     # if we're able to provide this via HathiTrust agreement we should discourage anyone from tyring to place a hold
     @holdable = nil if @document['oclc']
   end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -63,7 +63,7 @@ class CatalogController < ApplicationController
     @document['oclc'] = HathitrustOverlapRecord.find_by(catalog_id: @document['id'])&.oclc
 
     # if we're able to provide this via HathiTrust agreement we should discourage anyone from tyring to place a hold
-    @holdable = nil if @document['oclc']
+    @holdable = nil if @document['oclc'].present?
   end
 
   configure_blacklight do |config|
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -62,7 +62,7 @@ class CatalogController < ApplicationController
 
     @document['oclc'] = HathitrustOverlapRecord.find_by(catalog_id: @document['id'])&.oclc
 
-    # if we're able to provide this via HathiTrust agreement we should discourage anyone from tyring to place a hold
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -61,6 +61,9 @@ def show
     @document.delete('publisher_tesim') if @document['published_display']
 
     @document['oclc'] = HathitrustOverlapRecord.find_by(catalog_id: @document['id'])&.oclc
+  
+    # if we're able to provide this via HathiTrust agreement we should discourage anyone from tyring to place a hold
+    @holdable = nil if @document['oclc']
```
mbarnett, line 66: This is an absolutely tiny niggle, but maybe `if @document['oclc'].present?` just in case there's an empty string as the oclc value, for some reason?

PR #2055, opened 2020-09-14
Default translation

pgwillia:
## Context

We're not going to get upstream updates for Rails 4.2 and we've pinned blacklight so we've made the changes there and now we need to use them in discovery.

It looks like bundler 2.x is not compatible with Rails 4.2.  I'm not sure what combination of things gave me the impression that we could do the bump before.

Related to CVE-2020-15169 and https://github.com/ualbertalib/blacklight/pull/5

## What's New

Revert bundler to 1.17.3 and bump blacklight

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 
 ## [Unreleased]
 
+### Security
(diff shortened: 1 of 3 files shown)


murny (approved):
👍

pgwillia:
Sorry.  I pushed to my fork and Danger's not happy.  I'm going to close and recreated the PR.

Issue #2050, opened 2020-09-01
Add OCLC number (Marc 035a) to the search index.


redlibrarian:
This is an issue that has arisen with NEOS Discovery, but the two systems use the same search index. Concordia library uses WorldCat Local, and OCLC does not index our 001 cataloguing key. This means, they can't link directly to an item for hold-placing purposes. They *can* use the OCLC number, which *is* indexed in WorldCat Local, but we currently don't index that field. If we index that field, we may be able to come up with a solution for them to build a working place hold link. 

**Describe the solution you'd like**
Add something like **oclc_number_tesim = 035a** to the [[Symphony index properties]](https://github.com/ualbertalib/discovery/blob/master/config/SolrMarc/symphony_index.properties). Also add the oclc_number_tesim to the Solr search config (can use the same weighting as call number, I think). Then, when the next full reindex happens, records will become searchable by OCLC number in both UAL and NEOS Discovery.

**Describe alternatives you've considered**
We thought about adding this functionality to the place hold system itself, but we're limited by the Symphony API response, which can't search by anything but the 001/catkey.

pgwillia:
Is this only relevant to materials in the symphony ingest?  Not databases, kule or sfx?

redlibrarian:
That's right, yes. Only Symphony material gets reported to OCLC and gets an OCLC number. It's possible that there could be overlap with some Kule material, but this is mostly a solution to a NEOSDiscovery problem with placing holds, so that shouldn't be a concern.

PR #2058, opened 2020-09-14
added oclc number to symphony indexing

pgwillia:
## Context

Concordia library uses WorldCat Local, and OCLC does not index our 001 cataloguing key. This means, they can't link directly to an item for hold-placing purposes. They can use the OCLC number, which is indexed in WorldCat Local, but we currently don't index that field. If we index that field, we may be able to come up with a solution for them to build a working place hold link.

Related to #2050 https://github.com/ualbertalib/blacklight_solr_conf/pull/8

## What's New

index `035a` as `oclc_number_t`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Discovery project adheres to [Semantic Versioning](http://semver
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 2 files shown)


redlibrarian (approved):
This looks good to me. We will start to see the field become searchable as new records are added to the index (nightly) and then it will be fully in place next month when we do the next full reindex. I will do some test searching as this moves forward.

Issue #1875, opened 2019-12-20
Re-design: Improve error handling and reporting in ingest processes


nmacgreg:
**Resulting from a problem**
See: https://helpdesk.library.ualberta.ca/otrs/index.pl?Action=AgentTicketZoom;TicketID=55157
When the cron job for ingesting updates to SFX / eJournals data failed last night, it highlighted some limitations with this code and the environment surrounding it: 

1. The output of the cron/rake job is difficult to process _mechanically_
2. The job removes the existing data, before analysing whether the download of new, replacement was successful
While we have a monitoring solution (Nagios queries the data, to see if any is missing) to detect this situation, it should be possible to leave the existing data in place, while alerting the appropriate staff that the daily update process is failing. 

**Towards a solution**
1. The rake tasks for ingest need comprehensive, mature error handling. For instance, if the task fails to download the new data file, it should stop, report the problem, and not remove the old data, nor try to ingest a file that doesn't exist.
2. Whenever errors occur, clearly and consistently report errors, in a form that can be read both mechanically (eg by Nagios) and that makes it clear to operators how to resolve the problem.
3. When designing an improved solution, learn about the features of our monitoring system (or interface with experts), and take them into account, making it easy for Nagios to accurately pinpoint the problem and alert the staff.

**Additional Context**
Whether sfx/databases/symphony data, and whether run from the command line or a cron job, error detection, handling, and reporting should be consistent, and integrated with our monitoring system, so that operator attention can be efficiently directed at problems!

pgwillia:
From @nmacgreg's email "Re: missing journals"

Given the high impact of this problem to our patrons, the recent frequency of troubles, and the long, long time that these problems have existed in the codebase, I hope you set the priority of this issue very high!

Can I question design choices, again? I've brought this up before... there's probably still a GH issue...

- Why does the output of the job contain confusing, blatant error messages even when it works correctly?  This makes it hard to parse both by humans, and automation - eg:
```
INFO [main] (MarcImporter.java:749) -  Connecting to remote Solr server at URL http://solrcloud-prod.library.ualberta.ca:8080/solr/discovery-prod/update
    java.io.IOException: Server returned HTTP response code: 500 for URL: http://solrcloud-prod.library.ualberta.ca:8080/solr/discovery-prod/admin/registry.jsp
```
- Why is the actual error message not better highlighted, in the log output? (Again, making it hard to parse) - eg:
```
[...]

PR #2059, opened 2020-09-15
circuit breaker to SFX ingest

pgwillia:
## Context

Ocassionally we'll get a response from https://sfx-resolver.hosted.exlibrisgroup.com/resolver/cgi/public/get_file.cgi?file=sfxdata.xml that isn't utf-8 encoded and it ends up clearing out all our e-journals. Backing up the existing file and exiting if the new file isn't parsable should give us some more warning.

Also noticed some errors in the database_test_set so fixed that too.

Related to #1875 

## What's New

Fixed database_test_set and added circuit breaker for sfx ingest.

### DIFF
--- a/lib/tasks/ingest.rake
+++ b/lib/tasks/ingest.rake
@@ -46,7 +46,7 @@ task :ingest, [:collection] => [:update_solr_marc_maps] do |_t, args|
   @c = IngestConfiguration.new(args.collection, @config_file)
 
   # store a copy before it's replaced
-  File.rename(Rails.root.join(@c.path), Rails.root.join(@c.path + '.bak')) if @c.endpoint
+  File.rename(Rails.root.join(@c.path), Rails.root.join(@c.path + '.bak')) if @c.endpoint && File.exist?(Rails.root.join(@c.path))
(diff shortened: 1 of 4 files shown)


lib/tasks/ingest.rake
```diff
@@ -45,8 +45,22 @@ task :ingest, [:collection] => [:update_solr_marc_maps] do |_t, args|
   @collection = args.collection
   @c = IngestConfiguration.new(args.collection, @config_file)
 
+  # store a copy before it's replaced
+  File.rename(Rails.root.join(@c.path), Rails.root.join(@c.path + '.bak')) if @c.endpoint
```
mbarnett, line 49: Are there any cases where we'd need to first test for existence, before reopening (ie. does the process that generate this ever failed in a way such that the file never got created?). That's the only potential issue I can anticipate right now...

mbarnett (commented):
Are there any cases where

lib/tasks/ingest.rake
pgwillia: Good catch.  I can guard against if the file doesn't exist.

pgwillia: Does that work for you?

mbarnett (approved):
Sure! That works

Issue #2015, opened 2020-06-16
Rails 4.2 is unsupported release


pgwillia:
**Describe the bug**
When a release series is no longer supported, it's your own responsibility to deal with bugs and security issues. We [Rails] may provide backports of the fixes and publish them to git, however there will be no new versions released. If you are not comfortable maintaining your own versions, you should upgrade to a supported version. https://guides.rubyonrails.org/maintenance_policy.html

**Expected behavior**
We should be on a release series that is at least getting security patches. https://guides.rubyonrails.org/maintenance_policy.html

**Additional context**
Github is [warning us](https://github.com/ualbertalib/discovery/network/alert/Gemfile.lock/actionview/open) about a XSS vulnerability in ActionView.  Currently we're not affected but a fix would require a monkey patch rather than accepting a version bump.

pgwillia:
Rails 4.2 locks bundler <2.0

PR #2060, opened 2020-09-15
monkey patch rake CVE-2020-8181

pgwillia:
## Context

A directory traversal vulnerability exists in rack < 2.2.0 that allows an attacker perform directory traversal vulnerability in the Rack::Directory app that is bundled with Rack which could result in information disclosure.

We don't appear to use Rack::Directory in the application but monkey patching will ensure we're covered.

## What's New

monkey patch Rack::Directory#list_directory like https://groups.google.com/forum/#!topic/rubyonrails-security/IOO1vNZTzPA says but adapted for Rack 1.6-stable https://github.com/rack/rack/blob/47a1fd73fc77f094573f4215b0fc884f4ac1c03c/lib/rack/directory.rb#L79-L105

### DIFF
--- a/config/initializers/cve-2020-8161.rb
+++ b/config/initializers/cve-2020-8161.rb
@@ -1,0 +1,31 @@
+# https://groups.google.com/forum/#!topic/rubyonrails-security/IOO1vNZTzPA
+# but adapted for Rack 1.6-stable https://github.com/rack/rack/blob/47a1fd73fc77f094573f4215b0fc884f4ac1c03c/lib/rack/directory.rb#L79-L105
+
+Rack::Directory.prepend(Module.new do
+  def list_directory
+    @files = [['../','Parent Directory','','','']] 
+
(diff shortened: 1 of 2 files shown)


config/initializers/cve-2020-8161.rb
```diff
@@ -0,0 +1,32 @@
+# https://groups.google.com/forum/#!topic/rubyonrails-security/IOO1vNZTzPA
```
murny, line 1: Might be nice to throw a note here about when we can delete this? 

Once we upgrade to rack 2.2.0+ then this is no longer needed? Just seems rails 4.2 is pinning rack to this version. So once we upgrade Rails from 4.2 to 5.0 I think we can remove this I assume?

murny: Looks like you made a note about it in this issue 👍 
https://github.com/ualbertalib/discovery/issues/2015

murny (approved):

PR #2067, opened 2020-09-23
Circuit breaker memory

pgwillia:
## Context

`rake ingest[sfx]` after adding the circuit breaker caused the ingest script to run out of memory:
```
rake aborted!
Errno::ENOMEM: Cannot allocate memory - java
```
> Kilmarnock in Prod  has 4Gb memory / 2Gb swap and performs this feat monthly without complaint.
Forest has the same virtual h/w, but was throwing off Netdata alarms about memory & swap usage while ingesting SFX specifically

Basically we read the file twice with two different GC'd languages, thus the file's in memory twice.  oops

Related to #1875 

## What's New

Outsource the validation check to [xmllint](http://xmlsoft.org/xmllint.html) using Ruby library [Open3](https://devdocs.io/ruby~2.6/open3).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: ruby
 sudo: required
 addons:
   chrome: stable
+    apt:
+      packages:
+        # Needed for `xmllint`.
+        - libxml2-utils
 rvm:
   - 2.5
   - 2.6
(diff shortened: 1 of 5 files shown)


mbarnett (dismissed):
👍

pgwillia:
There was a problem with the travis config file.  Fixed now!

mbarnett (approved):
Alright, hopefully this does the trick now

PR #2069, opened 2020-09-29
fix memory use and bump for release

pgwillia:
## Context

After the last "fix" the large file continued to have issues.  Found the --stream flag for xmllint and it seems to have solved the mystery.

From `man xmllint`
```
--stream
Use streaming API - useful when used in combination with --relaxng or --valid options for validation of files that are too large to be held in memory.
```

Related to #1875

## What's New

use `--stream` flag when invoking `xmllint` on large sfx ingest files.


jefferya (approved):
LGTM!

PR #2082, opened 2020-10-15
skip these tests only on travis CI

pgwillia:
## Context

```
Failure/Error: visit
     Net::ReadTimeout:
     Net::ReadTimeout
```

The only change I could see between working and non working runs was the solr docker container and google chrome 86.

Related to https://travis-ci.org/github/ualbertalib/discovery/builds/732915914

## What's New

Skip the failing tests in the CI environment.  Also added a reminder to run tests in PR template.

PR #2083, opened 2020-10-15
Revert "skip these tests only on travis CI"

pgwillia:
Reverts ualbertalib/discovery#2082

maybe we shouldn’t skip those tests, then. sounds like there’s an actual non-CI issue here


pgwillia:
#2084 is a better resolution.

Issue #2121, opened 2020-12-17
Move from Travis CI to GitHub Actions CI tool


pgwillia:
The CI help run code tests within the GitHub code repository.

Motivation for the change of CI tools:

* alignment with other projects
* other projects have experienced a decrease in time to run tests
* https://travis-ci.community/t/extremely-poor-official-communication-of-the-org-shutdown/10568
* https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions

Similar to https://github.com/ualbertalib/avalon/issues/693

Issue #1932, opened 2020-03-03
Clean up migrations with seeding


pgwillia:
I guess my only concern is if a new developer joins this project and they need to create/migrate their local DB or anytime you blow away your current DB and start fresh (which I personally do every time I seed/migrate my DB which happens quite often) you will run these migrations which will seed the DB. This is because you start from a brand new DB and need to run every single migration since the projects inception. Which means you technically have to keep this up to date as well. So you will have to keep db/migrate/202002/circulation_rule.yml file in this repo forever and this seeding script in working condition or else this migration will fail.

Maybe why it's rare to see complicated seeding or other scripts in migration files as you want to keep migrations for simply schema changes or change existing data.

If it's a one time thing that we want for production, then maybe that should be a temporary rake task which gets ran once in production to seed the new tables. Then afterwards we can remove these files and the rake task and never have to think about this stuff again. If its not a one time thing and this is data that is required for the app (e.g you have a dropdown menu on a form populated from a DB table), then that should be in the seeds.

https://github.com/ualbertalib/NEOSDiscovery/pull/343#discussion_r385936595

I suppose the other option is leave these how they are. After they get ran in production, come back and cleanup all the scripts for seeding in these migrations and remove the yaml files? This could work too and removes my concerns. Plus could make Unix team job easier then having them run one time rake task

https://github.com/ualbertalib/NEOSDiscovery/pull/343#discussion_r385944385

PR #2125, opened 2020-12-22
Use github workflows to run CI

pgwillia:
## Context

With travis-ci transitioning to a limited credit system after Dec 31st and currently taking hours for jobs to go from queued to completion,  I think we'll need to migrate all our CI to github actions

Related to #2121

## What's New

- push workflow for github prepares and runs specs and rubocop


.github/workflows/push.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Lint with RuboCop
+        run: bundle exec rubocop --parallel
+
+      - name: Run Tests
+        env:
+          RAILS_ENV: test
+        run: |
+          bundle exec rake db:create 
+          bundle exec rake db:migrate
```
murny, line 44: Small tip, I think ` rails db:schema:load` is a bit more performant than doing migrations  (Will use the db/schema.rb to provision DB instead of running all the migrations). But thats totally up to you 👍

murny (approved):
Nice work 👍

.github/workflows/push.yml
pgwillia: 💯 You've definitely warned me about this before and I had hoped to fix it.  Right now there is some seeding that tests depend on that is set up by migrations.  It's a pain to deal with and I've learned my lesson.
https://github.com/ualbertalib/discovery/issues/1932

